PR #1, opened 2021-06-04
Noctua merges

cmpich:
Sierra, could you check if this contains all the changes you made on the ZFIN 1123 release branch? 
This is the difference between 1123 on ZFIN and 1123 on github.

### DIFF
--- a/server_apps/data_transfer/GO/go.pl
+++ b/server_apps/data_transfer/GO/go.pl
@@ -123,9 +123,9 @@ try {
 };
 
 try {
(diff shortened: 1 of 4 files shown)


sierra-moxon (approved):

sierra-moxon (changes requested):
the SQL for removing the annotations originally from ZFIN should be: 
```delete from zdb_active_data where zactvd_zdb_id not in (select mrkrgoev_zdb_id from marker_go_term_evidence where mrkrgoev_zdb_id = zactvd_zdb_id)```. Looks like this PR removes that qualifying statement?

PR #3, opened 2021-06-11
ZFIN-7062 Added timeout to GenBank Accession Update.

zfinryanm:
Tested on dev instance. (Using 'sleep 0.5h' added to command.)


cmpich (approved):
Looks good.

PR #4, opened 2021-06-29
ZFIN-7567 ZFIN-7057 Updated email properties for departure of Sabrina…

zfinryanm:
… and Prita. Also cleaned up Sierra, Patrick, and Kevin. Plus some other fixes.


zfinannee:
ZFIN-7538 on tracker is now ZFIN-7567 on zfin.atlassian.net

zfinryanm:
Trying this a different way that should be cleaner.

PR #14, opened 2021-08-30
ZFIN-7594 add GEO links

cmpich:


zfinryanm (commented):
Seems okay to me but I'm not sure I follow the logic completely since I don't know the code.

PR #21, opened 2021-09-03
ZFIN-7046 Changed directory path for copying solr index from prod.

zfinryanm:


zfinryanm:
This should be working now. You will need to delete the *_solr_data volumes and then start the compile container to recreate them with the proper files/permissions. (See the second commit.)

cmpich (approved):
Looks good.

PR #29, opened 2021-09-22
add 301 redirect

rtaylorzfin:
I added this redirect to address the issue 7610 for the warranty page.  One note: I normally would have tested this on cell.zfin.org, but I'm having some trouble getting the deployment to work.  It may be because I wasn't able to restart apache.  So I could use a little help with that workflow. Thanks!


cmpich (approved):
Looks good for the apache redirect. We still need to make the hyperlink change in source as well.

PR #32, opened 2021-09-28
Fix to self-signed cert to allow chrome to proceed

rtaylorzfin:
DOCKER:

Chrome will not let you proceed against the self-signed cert warnings unless the cert has some extended properties.  See this page:

https://eengstrom.github.io/musings/self-signed-tls-certs-v.-chrome-on-macos-catalina

and since that page lists the cert generate command without a CSR, I had to consult this to accomplish the same, but with a CSR flow:

https://security.stackexchange.com/a/176084


zfinryanm (approved):
Looks good to me.

PR #33, opened 2021-09-28
ZFIN-7561

rtaylorzfin:
Add a frontend validation check on creation of new construct.
Used the existing code in the jsp file as inspiration, but moved it over to the constructAdd.js file that
already had a validation function being called.

https://user-images.githubusercontent.com/90803397/135179810-39d327a3-8e05-4287-b91b-1c97a598e0d1.mp4


cmpich (approved):
Looks good.

PR #39, opened 2021-10-12
Spock Fixes

rtaylorzfin:
These changes fix some of the issues with the spock tests. In particular if a null pointer was thrown because an @Autowired bean wasn't available or the TemporaryFolder didn't get set up right, those issues should now be fixed.  There are still some remaining tests that are failing though.


cmpich (approved):
Looks good.

PR #43, opened 2021-10-15
ZFIN-7601

rtaylorzfin:
Add a null check so that if someone edits a feature to remove the start location and end location fields, it will not throw an exception.


rtaylorzfin:
ZFIN-7601

This null check allows deleting the location fields when editing a feature.

cmpich (approved):
Looks good.

PR #45, opened 2021-10-21
ZFIN-7667 replace ZDB-TERM ids with ECO term Ids in the Disease annotation file

cmpich:


cmpich:
Good catch. The clause was of no good use. Removed it.

PR #53, opened 2021-11-08
ZFIN 7692

rtaylorzfin:
Add feature to allow running gradle command that generates report of naming issues and potentially outputs SQL fixes:

 "gradle genotypeNamingIssues > /tmp/scratch2.csv"

Also add a liquibase file to fix 1746 of the simpler permutation issues.


source/org/zfin/nomenclature/repair/GenotypeFeatureNameComparator.java
```diff
@@ -0,0 +1,29 @@
...
+            return -1;
+        }
+        return String.CASE_INSENSITIVE_ORDER.compare(pattern1.toString(), pattern2.toString());
+    }
+
+    public static boolean listsEqual(List<GenotypeFeatureName> list1, List<GenotypeFeatureName> list2) {
+        if (list1.size() != list2.size()) {
+            return false;
+        }
+        for(int i = 0; i < list1.size(); i++) {
```
cmpich, line 22: Could use this library:
CollectionUtils.isEqualCollection(list1, list2)

rtaylorzfin: Good point. I replaced that function so as not to reinvent the wheel. (Had to use ListUtils.isEqualList though since order is important here.)

PR #54, opened 2021-11-09
Zfin 7707

rtaylorzfin:


package.json
```diff
@@ -35,7 +35,7 @@
     "imagesloaded": "^4.1.4",
     "immer": "^9.0.6",
     "jquery-bridget": "2.0.1",
-    "jquery-modal": "^0.9.2",
+    "jquery-modal": "0.5.9",
```
cmpich, line 38: What does the caret mean? Just curious.

rtaylorzfin: I was thinking that the caret would allow packages with minor version changes so I dropped it.  But after checking to be sure, it appears that it is more conservative on "0.x.x" versions.  So I guess we could've left it in. Though in our case it wouldn't matter since there's no 0.5.10, etc.:

https://medium.com/@Hardy2151/caret-and-tilde-in-package-json-57f1cbbe347b

server_apps/data_transfer/ResourceCenters/pullFromEZRC.pl
```diff
@@ -102,7 +102,7 @@ ($$)
     my $resourceCenter = $_[1];
 
     if  ($resourceCenter eq "EZRC"){
-	if (system("/local/bin/wget http://www.ezrc.kit.edu/downloads/$filename")) {
+	if (system("/local/bin/wget https://www.ezrc.kit.edu/downloads/$filename --no-check-certificate")) {
```
rtaylorzfin, line 105: Is the "--no-check-certificate" meant to be a permanent change, or do we want some type of reminder to go back and change it at some point?

rtaylorzfin (commented):
Looks good!

package.json
cmpich: Sounds good.

cmpich:
Also, it looks like you are picking up some changes from my branch. Could you include only your change in the PR?

server_apps/data_transfer/ResourceCenters/pullFromEZRC.pl
cmpich: Good question. I don't know. We may ask Ryan M. how to handle this in the longer run.

package.json
rtaylorzfin: Looks like I lied to you. There are versions 0.5.10 and 0.5.11. I was going from memory and thought we were at the latest in the 0.5.x series.

PR #58, opened 2021-11-10
ZFIN-7614

rtaylorzfin:
Uniprot load issues:

- Changed the perl script to read .dat files line-by-line instead of slurping up the entire file into memory (which was crashing the script).
- Fixed issue with checking uniprot database for invalid IDs (was scraping http://www.uniprot.org/uniprot/{id}, but always returned true)
- Added a somewhat silly "whirley" progress indicator
- Add some more output to indicate the current state of the script (even more would be nice). Including timestamps so we can track how long each step takes typically

### DIFF
--- a/server_apps/ZFINPerlModules.pm
+++ b/server_apps/ZFINPerlModules.pm
@@ -20,6 +20,9 @@ my %monthDisplays = (
     "Dec" => "12"
 );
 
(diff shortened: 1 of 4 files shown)


cmpich:
Looks good.

cmpich (approved):

PR #64, opened 2021-11-17
ZFIN-7703

rtaylorzfin:
Introduce service class.
Add related genes as direct attributions when an allele is added to a publication.
Add test to DbUnitTests.


rtaylorzfin:
Are there any cases where a feature is added to a publication, but we don't want to directly attribute related genes?

cmpich:
> Are there any cases where a feature is added to a publication, but we don't want to directly attribute related genes?

Not that I know of. You might ask a curator for this.

ybradford:
@rtaylorzfin I think we will always want to directly attribute related genes when a feature is added to a publication.

cmpich (commented):
Looks good.

PR #67, opened 2021-11-19
ZFIN-7735

rtaylorzfin:
ZFIN-7735

Reconfigure jenkins jobs to include 2 reports for finding possible issues in genotype names and fish names.


cmpich (changes requested):
You should add triggers for the two jobs in the trigger.production.properties file to schedule them to run weekly.

PR #75, opened 2021-12-04
ZFIN-7683 Add Timeout to Jenkins Jobs

rtaylorzfin:


cmpich (approved):
I would remove the test jenkins job. Otherwise, looks good.

PR #76, opened 2021-12-04
ZFIN-7628 fix certain data access to be compliant to JDK 17

cmpich:
ZFIN-7628 remove deprecated use in Java


rtaylorzfin (approved):
Looks good

PR #79, opened 2021-12-06
ZFIN-7735 Run Name Reports Daily

rtaylorzfin:
ZFIN-7735

Original pull request was configured to run weekly, but I think it makes sense to run them daily for now while we are still trying to find the source of the issues.  I also renamed Genotype-Name-Function-Mismatch to Check-Genotype-Name-Function-Mismatch to be consistent with the other reports.


cmpich (approved):
Looks good.

PR #80, opened 2021-12-07
Remove property files

cmpich:


rtaylorzfin (approved):
These changes all look good to me. I'm assuming the properties removed from buildfiles/allfilter.xml are no longer used anywhere.

PR #91, opened 2021-12-15
ZFIN-7758 Publication new layout

cmpich:


commons/env/schlapp-mac.env
```diff
@@ -24,3 +24,6 @@ export SOLR_PORT=9546
 export SOLR_MEMORY=8g
 export PGBINDIR="/Applications/Postgres.app/Contents/Versions/10/bin/"
 export SSH_USER="cmpich"
+export NODE_ENV="development"
+export ZFIN_ADMIN="cmpich@zfin.org"
+export WIKI_HOST="devwiki.zfin.org"
```
rtaylorzfin, line 29: What is devwiki.zfin.org?

home/WEB-INF/jsp/publication/publication-view-summary.jsp
```diff
@@ -0,0 +1,111 @@
...
+        <c:choose>
+            <c:when test="${!empty publication.keywords}">
+                ${publication.keywords}
+            </c:when>
+            <c:otherwise><span class="no-data-tag">none</span></c:otherwise>
+        </c:choose>
+    </z:attributeListItem>
+
+    <c:if test="${!empty publication.dbXrefs}">
+        <z:attributeListItem label="Microarrays">
```
rtaylorzfin, line 43: Should this be "datasets" (https://zfin.atlassian.net/browse/ZFIN-7737)

commons/env/schlapp-mac.env
cmpich: That is an old configuration parameter before we moved the wiki to the cloud. It is not used any longer and needs to be globally removed. Until then, it needs to be set.

rtaylorzfin (approved):
Looks good. I like the lombok getter/setter cleanups too.

PR #92, opened 2021-12-15
ZFIN-7704 Direct Attribution Cleanup

rtaylorzfin:
Includes a report generator that outputs an SQL fix.
Includes the latest SQL fix that was generated as liquibase file.
Only creates attributions if the publication is of "Journal" type.
Run the report command with:
    gradle  featureAttributionRepair
Report takes about 2 hours to run.


cmpich (approved):
looks good.

PR #95, opened 2021-12-20
ZFIN-7767

rtaylorzfin:
If deployPostgresFunctions fails then the build fails.
Clean up the function definitions that have syntax errors.

### DIFF
--- a/lib/DB_functions/build.gradle
+++ b/lib/DB_functions/build.gradle
@@ -21,24 +21,26 @@ int runSqlFiles(String directory) {
     def dbname = env['DBNAME']
     def pgbindir = env['PGBINDIR']
 
-
-    int numberOfProc = 0;
+    int numberOfProc = 0
 
     fileList(directory).each {
         File file ->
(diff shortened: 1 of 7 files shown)


cmpich (commented):
I need some explanation for the new SQL files.

rtaylorzfin:
Double check and make sure that before and after running deployPostgresFunctions, we get identical definitions of functions in the database.

Look into the dropFunctions build task to see why it is not dropping some functions

rtaylorzfin:
Just following up here. I confirmed that this commit will not change any existing database function. I did that by:

- dumping each of the database functions
- then running gradle deployPostgresFunctions
- then doing another dump of all database functions
- confirmed that they were the same.

As for where the build task for dropping functions is concerned, I'm not sure where that happens.  There seems to be a reference in Makefile to "gradle dropFunctions", but that's the only place I see it.  I'm not sure if the Makefile is ever used in the build process though?

cmpich (approved):

PR #98, opened 2021-12-27
Add skipNpmBuild flag

rtaylorzfin:
So we can do the following (for example):

  ant -DskipNpmBuild=1 dirtydeploy


rtaylorzfin:
Will use a different ant target instead.

PR #101, opened 2022-01-03
ZFIN-7765: Fix trigger for genotype names

rtaylorzfin:
- Change the trigger for feature_marker_relationship updates to happen after change (instead of before).
- Update naming issues repair console command to be able to run report and also make the fixes at the same time using Jenkins parameterized job.
- Allow passing the log4j2.configurationFile flag to console command with an example like the following:

     gradle -Dlog4j2.configurationFile=home/WEB-INF/log4j2.xml genotypeNamingIssues

- Add specific name updates to liquibase file for genotype names.
- Refactor get_genotype_display for readability -- extract large query into view, use pg_format for beautification.


rtaylorzfin:
Christian, when you have a sec, I'd like to hear your thoughts on the possible side-effects of this commit:

https://github.com/ZFIN/zfin/pull/101/commits/42eface3e6352e709452fa5fa0e91d6d85f27500

Perhaps we could come up with some specific tests to make sure there are no breaking changes.

cmpich (approved):

PR #106, opened 2022-01-05
ZFIN-7769 fix column / cell population of the stats table

cmpich:
Screen Shot 2022-01-05 at 6 56 42 PM


rtaylorzfin (approved):
Looks good

PR #108, opened 2022-01-06
ZFIN-7783 (release 1127): Fix Serialization Problem

rtaylorzfin:
Quick fix for serialization issue that happened as a result of the jackson upgrade.


cmpich (approved):
Looks good.

PR #111, opened 2022-01-11
ZFIN-7792 allow the dispatcher servlet to forward the OPTIONS request method

cmpich:


rtaylorzfin (commented):
This looks good to me, but do you want to remove the space from "ZFIN-779 2" in the commit message?

PR #113, opened 2022-01-11
ZFIN-7756: Update fish names with direct update statements

rtaylorzfin:
Change the updates to direct updates rather than using the p_update_related_fish_names function.
Ignore "Cooch Behar" fish_name in report of fish name problems.


cmpich (approved):
Looks good.

PR #114, opened 2022-01-12
ZFIN-7763: Use log4j2.xml file for less noisy gradle tasks

rtaylorzfin:
Introduce a log4j2.xml file for gradle tasks. Is this the best place to put the log4j2.xml file (home/WEB-INF/conf/agr/log4j2.xml)?


cmpich:
I would call the file just log4j2.xml and place it into 
home/WEB-INF/conf
directory.

rtaylorzfin:
Okay, moved the file location.

PR #115, opened 2022-01-12
ZFIN-7604: Add links to Single Cell Expression Atlas

rtaylorzfin:
Add an outgoing link from gene page to Single Cell Expression Atlas.
Use curation publication (ZDB-PUB-220103-3) as reference link.


cmpich (approved):
Looks good.

PR #116, opened 2022-01-13
Add imagemagick as requirement for tomcat container.

rtaylorzfin:
Tomcat container needs imagemagick for generating thumbnails.


cmpich (approved):
I am fine with it.

PR #122, opened 2022-01-25
ZFIN-7816: Frontend Mitigation (main branch)

rtaylorzfin:
Mitigation that handles the issue of getting empty objects back from the API a bit more gracefully.
The rendered table omits the links where there should be more data from the API, but it still
shows as much of the table as possible (instead of failing with console error message).

### DIFF
--- a/home/javascript/react/components/entity/EntityAbbreviation.js
+++ b/home/javascript/react/components/entity/EntityAbbreviation.js
@@ -15,24 +15,29 @@ const TYPE_CLASSES = [
 ];
 
 const EntityAbbreviation = ({entity}) => {
-    let className;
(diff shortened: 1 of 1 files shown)


rtaylorzfin:
Instead of rendering no table at all in this circumstance, this front-end change would render a table with some missing information.  For example, on foxa3 (/ZDB-GENE-980526-423#mutations), we would at least see this table:

Screen Shot 2022-01-25 at 2 44 50 PM

With a fix in place for the backend API, it should look like this:

Screen Shot 2022-01-25 at 2 44 42 PM

cmpich (approved):
Looks good.

PR #129, opened 2022-01-27
ZFIN-7817: Create new LabAttribution class as subclass of RecordAttribution

rtaylorzfin:
Fix issue when editing certain ESTs (eg. https://zfin.org/action/marker/marker-edit?zdbID=ZDB-EST-040130-186)

Added new subclass of RecordAttribution called LabAttribution for resolving the many to one relationship to a lab.

### DIFF
--- a/conf/hibernate.cfg.xml
+++ b/conf/hibernate.cfg.xml
@@ -101,6 +101,7 @@
         <mapping class="org.zfin.infrastructure.PersonAttribution"/>
         <mapping class="org.zfin.database.PostgresSession"/>
         <mapping class="org.zfin.infrastructure.PublicationAttribution"/>
+        <mapping class="org.zfin.infrastructure.LabAttribution"/>
         <mapping class="org.zfin.infrastructure.RecordAttribution"/>
         <mapping class="org.zfin.infrastructure.ReplacementZdbID"/>
         <mapping class="org.zfin.infrastructure.SourceAlias"/>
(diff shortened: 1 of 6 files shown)


rtaylorzfin:
This seems to fix the immediate issue. I think there still remains a question about what to do with clones that have a record attribution with a lab or with a person. I'll put more notes in the issue.

Some comments on the code that may need to be changed:

This code seems to be redundant (is there a way to have a single place to store this?):
1. http://github.com/rtaylorzfin/zfin/blob/2f2d8260d4b7149b475e34a9ab08892fb2250dde/source/org/zfin/infrastructure/RecordAttribution.java#L21-L21 
2. http://github.com/rtaylorzfin/zfin/blob/2f2d8260d4b7149b475e34a9ab08892fb2250dde/source/org/zfin/sequence.hbm.xml#L271-L271

IntelliJ is giving me a warning about this hibernate annotation:
http://github.com/rtaylorzfin/zfin/blob/2f2d8260d4b7149b475e34a9ab08892fb2250dde/source/org/zfin/infrastructure/LabAttribution.java#L15-L15
"Error:(17, 13) 'Many To One' attribute type should not be 'Lab'"
But it says the same thing about PersonAttribution, so I'll assume it's fine.

cmpich (approved):
Looks good! Thanks for looking into it.

rtaylorzfin:
The xml mapping code should be able to be removed:

https://github.com/rtaylorzfin/zfin/blob/2f2d8260d4b7149b475e34a9ab08892fb2250dde/source/org/zfin/sequence.hbm.xml#L256

rtaylorzfin:
We could submit a request to jetbrains to ask about the ManyToOne warning.

PR #132, opened 2022-01-28
ZFIN-7807: Create Jenkins Uniprot Preload Job

rtaylorzfin:
Create a jenkins job for uniprot preload. 
Clean up some of the output of the job to make it easier to read.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -39,8 +39,8 @@ services:
       POSTGRES_DB: zfindb
       POSTGRES_HOST_AUTH_METHOD: trust
     shm_size: 1g
-    ports:
-      - "127.0.0.1:5432:5432"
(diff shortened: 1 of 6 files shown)


cmpich (commented):
Is there an archive for the various report / error files? Will I be able to see them in Jenkins?

rtaylorzfin:
Good question. I create a flag that would "archive" the various files that get generated into a subfolder for troubleshooting purposes. Now that it's a Jenkins job, it would be good to convert that to the Jenkins way of handling artifacts. I can make those changes.

rtaylorzfin:
@cmpich The artifact archiving is now happening through jenkins.

cmpich (approved):
Looks good. 
PGHOST needs to be defined in the docker world.

PR #135, opened 2022-02-02
ZFIN-7638: Revert

rtaylorzfin:
Revert initial pull request for ZFIN-7638.  Since we don't have a good way to test this right now, it's probably best to just revert to the old version for now.  More info on the jira ticket.


cmpich (approved):
Looks good.

PR #136, opened 2022-02-08
ZFIN-7810 plus SCRUM-1162

cmpich:
The first issue is to create an Alliance AffectedGenomicModel object.
Second issues is to add chemicalOntologyId to the daf export file.


rtaylorzfin (approved):
Looks good

PR #147, opened 2022-02-14
ZFIN-7733 add expression link to Fish miRNA site

cmpich:


source/org/zfin/expression/service/ExpressionService.java
```diff
@@ -330,6 +332,15 @@ public MarkerExpression getExpressionForGene(Marker marker) {
         return markerExpression;
     }
 
+    public LinkDisplay getFishMiTRna(Marker marker, ForeignDB.AvailableName foreignDBName) {
```
rtaylorzfin, line 335: Should this be called "getFishMiRna"?

rtaylorzfin (commented):
This looks good to me. I left one comment about a function name that is "getfishmiTrna". Is that T meant to be there?  Also curious about the use of temporary tables. I know postgres supports temp tables that disappear after the session. That might be useful.

rtaylorzfin (approved):

PR #151, opened 2022-02-17
ZFIN-7803 Do not log to catalina.out

rtaylorzfin:
Log gradle scripts that extent AbstractScriptWrapper to stdout only (not catalina.out).
Check for alternate file location for uploaded file because on trunk, jenkins is setting WORKSPACE env var to the wrong path on trunk, instead using ROOT_PATH.

### DIFF
--- a/home/WEB-INF/log4j2.xml
+++ b/home/WEB-INF/log4j2.xml
@@ -44,7 +44,9 @@
         <Logger name="org.hibernate.engine.internal.StatefulPersistenceContext" level="error"/>
         <Logger name="org.zfin" level="warn"/>
         <Logger name="org.zfin.ontology.OntologySerializationService" level="info"/>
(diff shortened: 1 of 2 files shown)


rtaylorzfin:
This should (mostly) address the issue we ran into on the last test run.  However, there are still 2 lines that go to catalina.out:

2022-02-16 17:45:39,954 [main] WARN  internal.ConnectionProviderInitiator - HHH000022: c3p0 properties were encountered, but the c3p0 provider class was not found on the classpath; these properties are going to be ignored.
2022-02-16 17:45:40,039 [main] WARN  internal.DriverManagerConnectionProviderImpl - HHH000402: Using Hibernate built-in connection pool (not for production use!)

cmpich (approved):
Good.

PR #159, opened 2022-02-23
ZFIN-7803

rtaylorzfin:
Change console task to use different log4j2.xml file (conf/gradle-log4j2.xml)
and Log to STDOUT for gradle task.


cmpich (approved):
Looks good.

PR #160, opened 2022-02-24
ZFIN-7608 Fix bug with re-clicking ribbon cell to close it

rtaylorzfin:
Fix issue reported by Leyla about not being able to collapse a subject by re-clicking the current subject in the ribbon.


cmpich (approved):
Looks good.

PR #161, opened 2022-02-26
Upgrade the c3p0 driver and related files.

rtaylorzfin:
I was tired of seeing the below output in the AGR tasks. As an added bonus, it looks like the upgrade of c3p0 from 0.9.5.2 to 0.9.5.5 does address some CVEs.

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
2022-02-24 22:14:04,683 [main] WARN  internal.ConnectionProviderInitiator - HHH000022: c3p0 properties were encountered, but the c3p0 provider class was not found on the classpath; these properties are going to be ignored.
2022-02-24 22:14:04,689 [main] WARN  internal.DriverManagerConnectionProviderImpl - HHH000402: Using Hibernate built-in connection pool (not for production use!)
```


cmpich (approved):
Looks good.

PR #166, opened 2022-02-28
ZFIN-7771 Add evidence code to display of chromosomal location

rtaylorzfin:
Add the evidence code to the display mode of the "chromosomal location".

Also fix issue that was introduced when changing backing table and thus renaming "id" to "zdbID".  This was causing a console error.


cmpich (approved):
Looks good.

PR #168, opened 2022-03-02
ZFIN-7845 ZFIN-7737 ZFIN-7889

cmpich:
Added a column called 'Sequence' that displays the actual sequence data upon show/hide.
Also, reformatted MarkerEditChromosomalLocation to pass the lint code checker.


source/org/zfin/sequence/ForeignDB.java
```diff
@@ -196,22 +196,32 @@ public static AvailableName getType(String type) {
         }
     }
 
-    public boolean isZfishbook () {
+    public boolean isZfishbook() {
         if (displayName != null && displayName.equals("zfishbook"))
```
rtaylorzfin, line 200: This is a small tweak, but what do you think of using:

`"zfishbook".equals(displayName)`

instead of 

`displayName != null && displayName.equals("zfishbook")`

cmpich:
I refactored to use the corresponding enum instead.

rtaylorzfin (approved):
Looks good

PR #174, opened 2022-03-08
ZFIN-7622: Replace config pkalita with ryanm

rtaylorzfin:
ZFIN-7622

Replace pkalita with ryanm to match the dizzy and moonshine files in commons/env.


cmpich (approved):
Looks good.

PR #182, opened 2022-03-23
ZFIN-7907 - Code cleanup for uniprot AND perl debugging in docker container

rtaylorzfin:
I cleaned up the sp_check.pl file a little bit. I also added the perl debugging libraries for help during development.


cmpich (approved):
Looks ok to me...

PR #188, opened 2022-03-29
ZFIN 7891: Fix blank fish names

rtaylorzfin:
Updating queries to include checks for empty names.
Key the fixes in GenotypeNamingIssues by both the original name and the computed name.
Add Yvonne and Ryan T. to recipient list for Check-Genotype-Name report and Check-Fish-Name report failure notifications.
Add confirmation before deleting feature marker relationship.
Broaden the scope of the report that checks for incorrect fish names.
Add liquibase for fixing fish and genotype names for known issues.
get_fish_name uses current genotype name. Updated to use get_genotype_display() function if current genotype name is blank.


server_apps/DB_maintenance/validatedata/Check-Fish-Name-Function-Mismatch_d.sql
```diff
@@ -1,5 +1,20 @@
...
+    fish_zdb_id,
+    fish_name,
+    get_fish_name (fish_zdb_id) AS computed_fish_name
+FROM
+    fish
+WHERE (
+            fish_name <> get_fish_name (fish_zdb_id)
+        OR trim(fish_name) = ''
+        OR trim(get_fish_name (fish_zdb_id)) = ''
+        OR trim(fish_name) = '(AB)' -- basically a blank name, but with a background
```
cmpich, line 11: What about other background gentoypes, ie other than 'AB', e.g. 'TL'?

rtaylorzfin: Good point. I updated it to look for names that start with '('.

cmpich:
Looks good.

PR #191, opened 2022-03-31
ZFIN-7591: Working blast setup under docker. Need to download necessa…

zfinryanm:
…ry blast dbs and binaries to complete setup.


rtaylorzfin (approved):
Looks good to me. I checked out the branch, updated my .env, rebuilt docker images, and ran the gradle task successfully. My only comment would be if we should add anything to docker/environment_linux? I don't know if we're using that anywhere.

zfinryanm:
I don't think anyone is using environment_linux right now so I didn't bother updating it. It'll need to be fixed before we move to using docker on the dev VMs, I'd guess.

PR #199, opened 2022-04-14
ZFIN 7962: Allow jbrowse in main branch behind feature flag

rtaylorzfin:
This pull request allows you to pass a url parameter of "?jbrowse=true" on the end of a url for a page with gbrowse and it will switch to the jbrowse version of react component and html element.


cmpich:
With your approach you need to add the '?jbrowse=true' part to each URL, i.e. to each page you are looking at. I'd rather have a session parameter that would remember that I am in jbrowse land for all pages. Is that doable?

rtaylorzfin:
That sounds doable to me!

cmpich (approved):
Looks good.

PR #200, opened 2022-04-18
ZFIN-7936: Pub prototype copyable ID/pubmedID

rtaylorzfin:
Invert nesting of elements (from 'a > h5' to 'h5 > a').  Allow passing subsection or "addendum" to the entity name. Allows the secondary text to be "copyable".


cmpich (approved):
Looks good.

PR #202, opened 2022-04-22
ZFIN-7946

rtaylorzfin:
Add a curl command to hit the update-cache endpoint after generating downloads directory.


rtaylorzfin:
Yeah, it is a bit of a workaround. I'm not sure it will work, but worth trying. It's possible that the file watching code was working as expected, but the part that checks what new files exist downstream is not seeing the new files. If this turns out to fix the issue, I agree we should clean up any unnecessary file-watching code.

PR #203, opened 2022-04-23
(Placeholder) ZFIN-7813: Highlight string for jBrowse use location instead of ID

rtaylorzfin:
This PR has a bunch of commits because it was based on the ZFIN-7962 PR.  So let's not merge this in until that one is merged first. Then we may want to close this one and re-open it.


cmpich:
Can you re-create this PR when you merged the other jBrowse ticket into main? Then this ticket should have many fewer file changes.

PR #205, opened 2022-04-25
ZFIN-7953

rtaylorzfin:
Fix issue with setting properties for existing fish when creating a fish on the curation page.


source/org/zfin/mutant/repository/HibernateMutantRepository.java
```diff
@@ -1800,7 +1800,7 @@ public boolean createFish(Fish fish, Publication publication) {
         Fish existingFish = getFishByGenoStr(fish);
         boolean newFishCreated = false;
         if (existingFish != null) {
-            fish = existingFish;
+            fish.setFishProperties(existingFish);
```
cmpich, line 1803: This approach feels clunky... 
I'd rather re-fetch the record through Hibernate from the database (see my PR).

PR #207, opened 2022-04-26
ZFIN-7953 return persistent fish object 

cmpich:


rtaylorzfin (approved):
Looks good. Much simpler than mine!

PR #210, opened 2022-04-26
ZFIN-7813: Use Genome Location Information for jBrowse Highlighting

rtaylorzfin:
Translate highlight string for features and markers by using values from sequence_feature_chromosome_location_generated table via LinkageRepository::getGenomeLocation.


source/org/zfin/jbrowse/presentation/JBrowseImageBuilder.java
```diff
@@ -70,6 +75,39 @@ public GenomeBrowserImage build() {
...
+    }
+
+    private GenomeLocation getGenomeLocation(ZdbID markerOrFeature) {
+        List<GenomeLocation> genomeLocations = getLinkageRepository().getGenericGenomeLocation(markerOrFeature);
+
+        Optional<GenomeLocation> first = genomeLocations.stream().filter(
+                genomeLocation -> getGenomeBuild().getValue().equals(genomeLocation.getAssembly())
+        ).findFirst();
+
+        GenomeLocation result = null;
```
cmpich, line 104: Better to use the exception safe orElse() method instead of get(), e.g.
return first.orElse(null);

PR #217, opened 2022-05-04
ZFIN-7982 retrieve and presist reference proteins from UniProt

cmpich:


source/org/zfin/db/postGmakePostloaddb/1132/ZFIN-7982.sql
```diff
@@ -0,0 +1,17 @@
+--liquibase formatted sql
+--changeset cmpich:ZFIN-7982
+
+
+create table REFERENCE_PROTEIN
```
rtaylorzfin, line 5: Will this make the table name uppercase?

cmpich: I believe postgres is case-insensitive in table names, though, I try to adhere to the convention to capitalize table names.

rtaylorzfin (approved):
Looks good

PR #218, opened 2022-05-04
ZFIN 7932: Fix issues validating against alliance API

rtaylorzfin:
Validation was failing due to non-existence of zfin_genes.gff3 in the directory that contains all the other files we upload to the alliance.  I added a line that copies zfin_genes.gff3 from the downloads directory where it is generated to the alliance uploads directory.  I also added an explicit check for the existence of each file so a missing file error will be easier to understand.


cmpich (approved):
Looks good.

PR #219, opened 2022-05-05
(PREVIEW) AGP Load

rtaylorzfin:
This loads the clone AGP file into the database. Currently need to manually create the table (see comment in AGPEntry.java).
Run command like the following to load into DB:

AGP_FILE=./GRCz11.agp gradle importCloneAssembly

Some things I need to figure out:
- What to name the DB table?
- After importing this data, how to manipulate it into the sequence_feature_chromosome_location_generated table
- The data contains "chomosome_start/end" fields and also "fragment_start/end" fields. Not sure the distinction.

Here's a screenshot of DB after import:

Screen Shot 2022-05-04 at 4 58 11 PM

### DIFF
--- a/console.gradle
+++ b/console.gradle
@@ -23,5 +23,14 @@ task journalAbbreviationSync(type: JavaExec) {
     main = 'org.zfin.publication.JournalAbbreviationSyncTask'
 }
 
+task importCloneAssembly(type: JavaExec) {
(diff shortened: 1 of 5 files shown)


cmpich (commented):
This looks good so far. I am not sure where and how this is used in the UniProt load.
The name of the table needs to be agreed on. Need to understand the business logic of this file / data.

PR #221, opened 2022-05-05
Fix GBrowseImageSpec.groovy tests 

rtaylorzfin:
Use the new GenomeBrowserFactory method of creating GbrowseImageBuilder.


cmpich (approved):
Looks good.

PR #226, opened 2022-05-09
ZFIN-7962-hotfix (release-1132)

rtaylorzfin:
Add a null check for GBrowseImage in the react component container element. 
(I previously created two pull requests for main branch, instead of one against release branch.)


cmpich (approved):
Looks good.

PR #227, opened 2022-05-09
Extract a jsp tag for gbrowse/jbrowse

rtaylorzfin:
This is a cleaner way of rendering the standard gbrowse/jbrowse image without duplicating boilerplate code in JSPs.


cmpich (approved):
Looks good.

PR #229, opened 2022-05-10
ZFIN-7812: Load AGP file and translate into MarkerGenomeLocation entities

rtaylorzfin:
Load entries from AGP file (eg. /research/zprodmore/gff3/GRCz11.agp). 
Translate those entries into MarkerGenomeLocation objects to be used 
with JBrowse for displaying location of clones.


cmpich (approved):
Looks good. Just ask Ken if the additional location info is fine to have in our post gres database

PR #231, opened 2022-05-10
ZFIN-7741: UniProt Load - Handle merged genes gracefully

rtaylorzfin:
If the uniprot load comes across a record with 2 zfin genes associated with it 
and one of the genes has been merged into the other gene, then omit the old
gene information from that record.


cmpich (approved):
Looks good.

PR #233, opened 2022-05-13
ZFIN 8009: Fix GBrowseImageSpec Failure, other gbrowse regressions

rtaylorzfin:
Fix hibernate cast from "(Object)sources" to "(Object[])sources".

Fix gbrowse regression: GenomeLocation objects were not properly loading the gbrowseTrack field from the column in the database.


cmpich (commented):
Good.

cmpich (approved):
Looks good.

PR #234, opened 2022-05-13
ZFIN 8009 hotfix to release-1132

rtaylorzfin:
Fix hibernate cast from "(Object)sources" to "(Object[])sources".

Fix gbrowse regression: GenomeLocation objects were not properly loading the gbrowseTrack field from the column in the database.


cmpich (approved):
Looks good.

PR #235, opened 2022-05-14
Hotfix for Hibernate Upgrade bug

rtaylorzfin:
Seems like "fetch all properties" is not having any effect.
We are getting a hibernate error when ```            newAntibodyStat.addFigure(figure);```
is called because it is resolving a lazy-loaded relationship and this is
happening while in the context of iterating over a ScrollableResults set which doesn't
allow running queries. Must be due to hibernate upgrade.

### DIFF
--- a/source/org/zfin/expression/repository/HibernateExpressionRepository.java
+++ b/source/org/zfin/expression/repository/HibernateExpressionRepository.java
@@ -1803,7 +1803,7 @@ public class HibernateExpressionRepository implements ExpressionRepository {
                 "        and fishox = xpExp.fishExperiment " +
                 "        and xpRslt.expressionExperiment = xpExp " +
                 "        and xpExp.gene != null" +
-                "        and xpExp.gene.zdbID not like :markerId)";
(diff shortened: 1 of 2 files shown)


source/org/zfin/expression/repository/HibernateExpressionRepository.java
```diff
@@ -1803,7 +1803,7 @@ public List<ExpressionResult> getNonEfgExpressionResultsByFish(Fish fish) {
                 "        and fishox = xpExp.fishExperiment " +
                 "        and xpRslt.expressionExperiment = xpExp " +
                 "        and xpExp.gene != null" +
-                "        and xpExp.gene.zdbID not like :markerId)";
+                "        and xpExp.gene.zdbID not like :markerId";
```
rtaylorzfin, line 1806: I don't understand how this ever worked in the past?

cmpich: Good find.

cmpich (approved):
Looks good. Should find example fish page that should display non-EFG expression data.

PR #237, opened 2022-05-16
Fix antibody section on term detail page due to hibernate upgrade: /a…

cmpich:
…ction/ontology/show-labeled-antibodies


rtaylorzfin (approved):
Looks good

PR #238, opened 2022-05-16
Release 1132 GBrowse Range Hotfix

rtaylorzfin:
Fix issue where the range was calculating incorrectly because the function call "withPadding(10000)" was calling the function with signature of "withPadding(double d)" instead of "withPadding(int i)".  Turned out to be due to the integer one not being defined on the new interface.

Screenshots showing the regression:

zfin.org:
Screen Shot 2022-05-16 at 12 47 02 PM

trunk:
Screen Shot 2022-05-16 at 12 47 11 PM


rtaylorzfin:
I'm going to merge this in so I can deploy on test.zfin.org.  Leaving the PR for main branch around for review.

PR #239, opened 2022-05-16
ZFIN-8009: GBrowse Range Hotfix

rtaylorzfin:
Fix issue where the range was calculating incorrectly because the function call "withPadding(10000)" was calling the function with signature of "withPadding(double d)" instead of "withPadding(int i)". Turned out to be due to the integer one not being defined on the new interface.

Screenshots showing the regression:

zfin.org:
Screen Shot 2022-05-16 at 12 47 02 PM

trunk:

Screen Shot 2022-05-16 at 12 47 11 PM


rtaylorzfin:
Thanks for looking it over. I decided to add the overrides when I realized that I might have avoided this problem altogether if I had started with override annotations. Better late than never, I suppose.

PR #240, opened 2022-05-17
ZFIN-8011: Fix hibernate issue from test suite

rtaylorzfin:
Call getAllPublicationsForFeature at top of method to avoid throwing an exception 
(HibernateException: identifier of an instance of org.zfin.profile.FeatureSupplier 
was altered from org.zfin.profile.ObjectSupplier$HibernateBasicProxy$eAKO0ILG@369902662 
to org.zfin.profile.FeatureSupplier@5cc6fc3)

More details in the ticket: https://zfin.atlassian.net/browse/ZFIN-8011


cmpich (approved):
Looks good. I am not sure why moving the line is solving the issue.

PR #241, opened 2022-05-17
ZFIN-7638 Fix Check-And-Update-Journals_w to not overwrite abbrevs with null

rtaylorzfin:
SQL statement in Check-And-Update-Journals_w was not correctly joining tables so it would overwrite abbreviations.


cmpich (approved):
Looks good.

PR #243, opened 2022-05-22
ZFIN-7977: Remove dependency on faceted search from uniprot load

rtaylorzfin:
This preserves IDs from entries in the marker_go_term_evidence table so that when they are reloaded, they don't get new IDs that require indexing to run again.


cmpich (approved):
Looks good.

PR #248, opened 2022-05-26
ZFIN-8019: Fix xml serialization error

rtaylorzfin:
Not sure if it's important to retain the fork="false" attribute. The jvmargs don't take effect unless we fork though.


cmpich (approved):
Looks good.

PR #252, opened 2022-06-02
ZFIN-7964 add release version name in header

cmpich:


source/org/zfin/framework/presentation/ZfinJSPFunctions.java
```diff
@@ -409,4 +410,11 @@ public static String getAssetPath(String name) {
 
     }
 
+    public static String getSoftwareBranch() {
+        String softwareVersion = VersionService.getSoftwareVersion();
+        if (softwareVersion.startsWith("release-")) {
```
rtaylorzfin, line 415: Is there a possibility of null value here for softwareVersion?

cmpich: Good point. Added 'unknown' value in case no git info is available.

PR #258, opened 2022-06-03
ZFIN-7977 (release-1133): Replace protein_db_link relationship with uniprot accession number column

rtaylorzfin:


cmpich (approved):
Looks good.

PR #259, opened 2022-06-03
NPE thrown

rtaylorzfin:
I deployed in my dev environment without running deployGitInfoFile and got a null pointer because inputStream was evaluated as null. Probably wouldn't happen on production.


cmpich (approved):
Looks good.

PR #264, opened 2022-06-08
ZFIN-7945 and ZFIN-7944

rtaylorzfin:
For NTR pages, this allows adding citations to the "chromosomal location" information.  It also displays the same information to the end users on NTR view pages.

### DIFF
--- a/source/org/zfin/marker/repository/HibernateMarkerRepository.java
+++ b/source/org/zfin/marker/repository/HibernateMarkerRepository.java
@@ -59,10 +59,10 @@ import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
 import javax.persistence.criteria.Root;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static org.zfin.framework.HibernateUtil.currentSession;
-import static org.zfin.repository.RepositoryFactory.getInfrastructureRepository;
-import static org.zfin.repository.RepositoryFactory.getMarkerRepository;
+import static org.zfin.repository.RepositoryFactory.*;
 
 
 @Log4j2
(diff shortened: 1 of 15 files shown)


rtaylorzfin:
@cmpich , there are a couple places I'd appreciate your feedback:

1. I am introducing the isMarkerNtr variable to pass to the jsp for conditional rendering. I'm not sure if this is the most elegant way to do this.
2. I am storing citations for these "chromosomal locations" as RecordAttribution objects. I think this is the right way to go, but just wanted to check with you.  The entries would have a data_zdb_id that corresponds to the sequence_feature_chromosome_location table and a source_zdb_id to the publication.

source/org/zfin/marker/repository/HibernateMarkerRepository.java
```diff
@@ -793,6 +794,57 @@ public void addDataAliasAttribution(DataAlias alias, Publication attribution, Ma
...
+
+            infrastructureRepository.insertUpdatesTable(genomeLocation.getZdbID(), "", "new attribution publication to genome location " + attributionZdbID + " to " + relZdbID);
+        }
+
+    }
+
+    public void synchronizeGenomeLocationAttributions(MarkerLocation genomeLocation, Set<String> publicationIDsToSync) {
+        //ADD
+        for (String publicationID : publicationIDsToSync) {
+            this.addGenomeLocationAttribution(genomeLocation, publicationID);
```
cmpich, line 831: do you need the 'this.' prefix to call the method in this class?

rtaylorzfin: Good point

cmpich (approved):
Is there no general controller method from the other UI components that creates an attribution for an entity, in your case a Location?

rtaylorzfin:
> 

That's a good suggestion.  I looked through the code and didn't see anything that seemed like the right fit.

PR #266, opened 2022-06-09
ZFIN-7994 create Jenkins Job for Reference Proteome Load

cmpich:
the two SQL files are only capturing the queries for the reports requested in the two jira tickets. They will not be executed by any conmand.


rtaylorzfin (approved):
Looks good

PR #268, opened 2022-06-10
Persistent flags

rtaylorzfin:
Store feature flags in the database so they can be set globally for all users.


source/org/zfin/infrastructure/repository/HibernateInfrastructureRepository.java
```diff
@@ -686,6 +687,21 @@ public ZdbFlag getUpdatesFlag() {
         return (ZdbFlag) criteria.uniqueResult();
     }
 
+    public FeatureFlag getFeatureFlag(String name) {
+        Session session = HibernateUtil.currentSession();
+        org.hibernate.query.Query query = session.createQuery("from FeatureFlag ff WHERE ff.name = :name");
```
cmpich, line 692: I don't think you need the alias 'ff'...

rtaylorzfin: Good point. Removed.

cmpich (approved):
Looks good.

PR #270, opened 2022-06-15
ZFIN-7996 ZFIN-8045 ZFIN-8058 add prefix to curies in allele file, add tx for accession deletiong, correct error field name

cmpich:


rtaylorzfin (approved):
Looks good

PR #272, opened 2022-06-17
ZFIN-7925: Fix Failing NCBI Jenkins Load

rtaylorzfin:
Add a gradle task to perform the efetch of NCBI data as a replacement for /private/bin/efetch.r.
Update the NCBI_gene_load.pl file to use the new gradle task.
Remove deprecation issues in NCBIRequest.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -182,6 +182,7 @@ def translateVariableSet = [
         'GO_EMAIL_CURATOR',
         'GROOVY_CLASSPATH',
         'INSTANCE',
+        'JAVA_HOME',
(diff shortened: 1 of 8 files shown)


rtaylorzfin:
There was one thing on this PR where I was wondering if there is a better way.  On my first attempt at fixing, the jenkins job failed because gradle was trying to use the jenkins JAVA_HOME (jdk11).  I added a transform variable so the deployment will replace  in the perl script with the one in zfin.properties.  It works, but I'm curious if you've run across this type of thing before and have a better approach?

PR #274, opened 2022-06-20
Irrigon

cmpich:


source/org/zfin/marker/agr/GeneLinkMLInfo.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public GeneLinkMLInfo(int number) {
+        numfOfRecords = number;
+    }
+
+
+    public static void main(String[] args) throws IOException {
+        int number = 0;
+        if (args.length > 0) {
+            number = Integer.valueOf(args[0]);
```
rtaylorzfin, line 36: Where does args[0] get set? I remember gradle is weird about passing along command line arguments. Also, I think args[1] is the first argument and args[0] is the command being executed?

rtaylorzfin (approved):
Looks good

PR #279, opened 2022-06-29
ZFIN-8057: Fix external links that really on popovers.

rtaylorzfin:
Drop the positionFixed:True config option since it seems to no longer be supported.
https://getbootstrap.com/docs/4.1/components/popovers/

Convert class to lombok-style getters and setters.


zfinryanm (approved):
Looks okay to me.

PR #281, opened 2022-06-29
ZFIN-8059: Return images for publication by figure orderingLabel

rtaylorzfin:
Add ordering to image fetch for publication.

Convert class to lombok-style getters and setters.


rtaylorzfin:
Added a couple of fixes to issues I noticed on the publication page:

- STRs and Disease sections were in different order on the left navigation and the body of the page
- React was warning about multiple rows having the same keys:


Screen Shot 2022-06-29 at 11 06 05 AM

cmpich (approved):
Looks good.

PR #282, opened 2022-06-29
ZFIN-8063: Skip alliance submit for disease annotation without fish

rtaylorzfin:
Only submit disease annotation to alliance if it includes a fish.

If there is no fish on a disease annotation, then it is more like a regular attribution between a disease and publication. See ticket notes.


cmpich (approved):
Looks good.

PR #283, opened 2022-06-30
ZFIN-8077: Add a diagnostic page to devtools that shows request headers.

rtaylorzfin:
This doesn't fix the ultimate source of the issue, but it could be a helpful diagnostic. Example screenshot:

Screen Shot 2022-06-30 at 8 52 56 AM

### DIFF
--- a/source/org/zfin/framework/presentation/DatabaseInfoController.java
+++ b/source/org/zfin/framework/presentation/DatabaseInfoController.java
@@ -21,10 +21,7 @@ import javax.naming.InitialContext;
 import javax.naming.NamingException;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
-import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.lang.management.ManagementFactory;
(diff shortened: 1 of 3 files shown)


source/org/zfin/framework/presentation/DatabaseInfoController.java
```diff
@@ -21,10 +21,7 @@
 import javax.naming.NamingException;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
-import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
```
zfinryanm, line 27: Should these have been removed?

zfinryanm (commented):
Looks good other than the few lines I commented on below.

source/org/zfin/framework/presentation/DatabaseInfoController.java
rtaylorzfin: It should be safe to do so. They are not used in the file. I usually remove unused imports when IntelliJ points them out to me just to leave the file a little cleaner than when I found it.

zfinryanm (approved):

PR #289, opened 2022-07-11
(Release-1133) ZFIN-7945: One more int to long conversion

rtaylorzfin:
ZFIN-7945: One more int to long conversion


cmpich (approved):
Looks good.

PR #296, opened 2022-07-14
ZFIN-7991 ZFIN-8079 remove negative indentation for lab member list, fix table layout for non-paginated DataProvider tables

cmpich:


rtaylorzfin (approved):
Looks good

PR #297, opened 2022-07-15
Release-1133 Hotfix

rtaylorzfin:
Hotfix to address missing rows in gene ontology table on gene page.


cmpich:
Is this a one-time thing? How did these 'orphans' get into this state?

cmpich (approved):

rtaylorzfin:
I was thinking that it was a one-time thing. However, it does look like they get recreated.  I have another related ticket in this sprint that will give me a chance to investigate more. I highlighted the data that was lost from the gene ontology table on pax9:

Screen Shot 2022-07-15 at 7 59 56 AM

That gives me a clue that it has to do with loading the interpro association.

PR #303, opened 2022-07-21
ZFIN-8052, ZFIN-8113: Duplicates in gene_association.zfin.gz

rtaylorzfin:
Fix some issues with duplicate entries in gene_association.zfin.gz. The primary issue seems to be the sort order in the intermediary go.zfin file.


rtaylorzfin:
Change the jenkins job to an ant task so we don't clutter jenkins with jobs that curators need to sort through

cmpich (approved):
Looks good.

PR #304, opened 2022-07-22
ZFIN-7886: API to create a new sequence for marker

rtaylorzfin:
Add an API endpoint for creating a new sequence associated with a marker.


cmpich (approved):
Good.

PR #306, opened 2022-07-25
ZFIN-8115: Fix failing tests

rtaylorzfin:
Fix failing tests.
- Uniprot changed their API
- Mapping test depended on results to return in certain order
- Http requests to uniprot need to be fully read before closing connection
- Tests depended on duplicate marker_go_term_evidence entries we will be removing


cmpich (approved):
Looks good.

PR #321, opened 2022-08-11
Main Branch: Patch to fix angular deprecation with upgrade.

rtaylorzfin:
The $onInit function needs to be called so that all bindings are set up correctly before executing the function.


cmpich (approved):
Looks good.

PR #324, opened 2022-08-15
ZFIN-8151 ZFIN-8150 add non-genes to Genes/Marker table on publication page, fix sequence section on gene page

cmpich:


zfinryanm (approved):
Looks good.

PR #325, opened 2022-08-15
ZFIN-8154: Fix UniProt links

rtaylorzfin:
Fix UniProt links out and retain fixes to the UniProt tests for UniProt accession validation.


cmpich (approved):
Looks good.

PR #335, opened 2022-08-25
ZFIN-7925: Fix issues with NCBI Load 

rtaylorzfin:
This fixes issues with the NCBI Load not running and also the issue in ZFIN-7925 about the same accession number not getting associated with multiple genes.

Potential Issue: This deletes from reference_protein because it was causing foreign key violation errors.


cmpich (approved):
Looks good.

PR #336, opened 2022-08-25
UniProt: Fail gracefully, other tweaks

rtaylorzfin:
- Make sure command execution failure results in exiting script with error code.
- Use consistent casing
- Replace  with $ENV{'VARIABLE'}
- Fix this error:

```
psql:sp_load.sql:403: ERROR:  column "mrkrgoev_annotation_organization" is of type bigint but expression is of type text
LINE 5:          now() as time1, now() as time2, '5' as org, now() a...
```


cmpich (approved):
Looks good.

PR #338, opened 2022-09-04
ZFIN-8164  refactor MarkerRepository to take care of deprecated queries

cmpich:


source/org/zfin/mapping/MappingService.java
```diff
@@ -144,6 +149,17 @@ private static List<GenomeLocation> sortAndFilterGenomeBrowserLocations(List<? e
         }
         return finalGenomeList;
     }
+
+    public static List<MarkerLocation> getMarkerLocation(String zdbID) {
+        Session session = HibernateUtil.currentSession();
```
rtaylorzfin, line 154: I noticed this is still using the old version of org.hibernate.Criteria.  I assume that you are saving that for later as we make our way through the various repositories?

rtaylorzfin (approved):
Looks good! This is an impressive refactor! 

I noticed MappingService is still the deprecated Criteria, but I assume we'll get to that later.

PR #344, opened 2022-09-06
ZFIN-7925 (revision 2): Preserve existing records

rtaylorzfin:
- Added a preservation step for record_attributions that refer to a db_link that is preserved.  Previously, the db_link would be preserved, but the record_attribution was not preserved.
- Include debug files and prepareLog in the artifacts.
- More logging, including the new genes associated with multiple accession numbers. 
- Kick off the reference protein job after running NCBI Gene Load job. 
- Allow running in the old mode with flag (useful for comparing different results).


cmpich (approved):
Looks good.

PR #349, opened 2022-09-09
ZFIN-8167: record_attribution timestamps

rtaylorzfin:
Add columns:
 - recattrib_modified_at
 - recattrib_created_at
 - recattrib_modified_count

I think it's up for discussion whether or not we want a "modified count" or a "created at".  I thought the count could be useful for tracking if any rows get changed multiple times.  The "created at" column could potentially be useful for diagnostics too.


cmpich (approved):
Looks good.

PR #351, opened 2022-09-12
Release-1134 Patch for ZFIN-8167: Make transactions read-only

rtaylorzfin:
Make transaction read-only


cmpich (approved):
We'll give it a try.

PR #352, opened 2022-09-12
ZFIN-8188: Fix unmatched closing div tag in publication abstract

rtaylorzfin:
Fix unmatched closing div tag in publication abstract.


source/org/zfin/db/postGmakePostloaddb/1135/ZFIN-8188-remove-unmatched-closing-div-tag.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+--changeset rtaylor:ZFIN-8188
+
+select zdb_id from publication where pub_abstract ilike '%</div>' and pub_abstract not ilike '%<div%';
```
cmpich, line 4: Does this assume that we only encounter </div> with <div> elements but not vice versa? Is that actually true?

cmpich (approved):
Ok, all cases have only a </div> without a beginning tag. Then this script should fix them.

PR #356, opened 2022-09-14
ZFIN-8169 ZFIN-8186 ZFIN-8155 add external note to AB genotype, fix attribute name for json file to Alliance, suppress transcript from jenkins report

cmpich:


source/org/zfin/db/postGmakePostloaddb/1136/ZFIN-8169.sql
```diff
@@ -0,0 +1,13 @@
+--liquibase formatted sql
+--changeset cmpich:ZFIN-8169
+
+create temp table tmp_id as
+select get_id('EXTNOTE')
+from single;
```
rtaylorzfin, line 6: Is "from single" necessary? Not a big deal, just caught my eye.

rtaylorzfin (approved):
Looks good

source/org/zfin/db/postGmakePostloaddb/1136/ZFIN-8169.sql
cmpich: Correct. It is not needed. That was the syntax with Informix.

PR #357, opened 2022-09-14
ZFIN-8179: Lab page alignment

rtaylorzfin:
Fix wrapping of dt values and dd values in lab and company pages. Fix display of long company and lab names so they use ellipses.

- add option to pass an additional body class tag for styling specific pages
- add option to adjust column width of dt/dd pairs in attributeListItem.tag


cmpich (approved):
Looks good.

PR #371, opened 2022-09-29
Fix small typo

rtaylorzfin:


cmpich (approved):
I am not sure if the short notation won't do the trick. It does compile.

PR #373, opened 2022-10-03
ZFIN-8093:

rtaylorzfin:
Do not aggregate sequence dblinks by accession on the marker edit page (eg. https://cell.zfin.org/action/marker/gene/prototype-edit/ZDB-GENE-090313-318).  Instead aggregate by combination of accession and foreign_db.

Also, if deleting a db_link, delete the associated link in the reference protein table.


cmpich (approved):
Looks good.

PR #380, opened 2022-10-11
ZFIN-8242: Wrap overflow of the title in the left navigation for construct pages.

rtaylorzfin:
Wrap overflow of the title in the left navigation for construct pages.


home/css/datapage.scss
```diff
@@ -373,3 +373,10 @@ table.lab-view-summary-table dd a, table.company-view-summary-table dd a {
     white-space: normal; /* <-- remove for ellipses truncation */
 }
 /* End of genotype page formatting */
+
+/* Construct page formatting */
+.construct-view a.back-to-top-link {
```
cmpich, line 378: Can we not reuse the css class that was used for the genotype case?

cmpich (approved):
Looks good!

PR #381, opened 2022-10-12
ZFIN-8222 ZFIN-8041 include inferredGenes and inferredAlleles in DA file, remove Vega link in edit mode

cmpich:


rtaylorzfin (approved):
Looks good

PR #391, opened 2022-10-25
Feature flag JSP tweak and example use case

rtaylorzfin:
Add some tweaks to make the feature flags easier to use:
1. Create a jsp function to allow wrapping html elements in a feature flag.
2. If a flag exists as an enum, but not in the DB, default to false and insert the DB value if toggled to true.


home/WEB-INF/jsp/infrastructure/home/home-about-box.jsp
```diff
@@ -1,6 +1,14 @@
 <%@ include file="/WEB-INF/jsp-include/tag-import.jsp" %>
+<%@ page import="org.zfin.framework.featureflag.FeatureFlagEnum" %>
 
 <z:section title="About ZFIN">
+    <!-- Example Feature Flag Use -->
+    <c:if test="${zfn:isFlagEnabled(FeatureFlagEnum.CURATOR_JOB_POSTING)}">
+        <p class="lead">
+            <a href="https://zfin.atlassian.net/wiki/spaces/jobs/pages/2996503058/ZFIN+Jobs">
```
cmpich, line 8: this is the line that may need change every time we wanted to display this announcement as we may point to different wiki pages. I am not absolutely sure. If that would always be the same it would just work nicely.

cmpich: Yvonne confirmed that the wiki page for the curator job will always be the same wik page. So no concern then any more.

cmpich (approved):
Looks good.

PR #392, opened 2022-10-25
ZFIN 8231: Report for Thisse Data Matching

rtaylorzfin:
A java wrapper around a bunch of sql queries for matching thisse data to our existing data.
Gradle task for running it ("gradle thisseLegacyImportAnalysisTask").
Bumped the sourceCompatibility and targetCompatibility to 15 for the use of Text Blocks.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,8 +15,8 @@ apply from: 'console.gradle'
 
 //apply from: file('setupTestsReports.gradle')
 
-sourceCompatibility = 14
-targetCompatibility = 14
+sourceCompatibility = 15
+targetCompatibility = 15
 
 //apply from : '../zfinGradle/build.gradle'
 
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -15,8 +15,8 @@ apply from: 'console.gradle'
 
 //apply from: file('setupTestsReports.gradle')
 
-sourceCompatibility = 14
-targetCompatibility = 14
+sourceCompatibility = 15
```
cmpich, line 18: Shouldn't we use '17'

cmpich (approved):
Looks good...

build.gradle
rtaylorzfin: Good point. We probably should. I think in this case, it was so I could use Text Blocks.

PR #394, opened 2022-10-27
ZFIN-8201: Thisse Data Import

rtaylorzfin:
Add the rest of the files that Thisse sent us as imports to the thisse schema tables.


cmpich (approved):
Looks fine.

PR #395, opened 2022-10-29
ZFIN-8280: Publication Navigation Counts

rtaylorzfin:
This is an experiment to get the data counts to show up in the left navigation on the publications page.
I tried using redux, but the biggest drawback is that all components need to be in the same tree, which would require a fair amount of refactoring.
Instead, I created a simple storage object with listeners.
The downside so far is that the left navigation loads a bit more slowly.  We could maybe just use vanilla js instead of react components for the navigation.


rtaylorzfin:
Screenshot:

Screen Shot 2022-10-28 at 4 54 32 PM

PR #396, opened 2022-10-29
ZFIN-7933: solr configuration change (splitOnNumerics)

rtaylorzfin:
Add configuration splitOnNumerics="0" so that something like ENSDARG000000037177 doesn't get split into "ENSDARG" and "000000037177".


cmpich (approved):
Nice find! Looks good.

PR #397, opened 2022-11-01
ZFIN-8266: Fix construct search (example fli1:Dsred)

rtaylorzfin:
Fix constraint to reference construct table instead of fluorescent_marker table so that search returns constructs even if they don't have relationships to fluorescent markers.


cmpich:
Looks good!

PR #403, opened 2022-11-08
ZFIN-8280: Publication Navigation Counts

rtaylorzfin:
Show numbers next to the navigation items on the left side of publication page. These numbers are updated as each react component for the various sections loads. There is a flag for the dataPage jsp tag called "useNavigationCounter" that can be set to true in order to use the react components on the left hand nav for displaying section counts, otherwise, it reverts to the old version without any react.

Add DISTINCT to sql query for directly attributed ids.

Use feature flags so make this behavior togglable for publications view page.

Convert the figures display section to a react component. (though it isn't very "reacty").


cmpich (approved):
Looks good.

PR #409, opened 2022-11-09
1137 Hotfix: ZFIN-8293 - getABRegID

rtaylorzfin:
The former version of getABRegID had a "limit 1" clause.
That is no longer supported, so this change duplicates the same behavior.

Using a Text Block with pg_format style hql.


cmpich (approved):
Looks good.

PR #413, opened 2022-11-11
ZFIN-8251 ZFIN-8253 fix CuratedBlastDatabase regneration, handle secondary terms in GAF load.

cmpich:


rtaylorzfin (approved):
Looks good

PR #416, opened 2022-11-14
ZFIN-8293: Hotfix for Main Branch v2

rtaylorzfin:
another hotfix to handle the empty results case for uniqueResult

This is a fix to the fix for ZFIN-8293. The first one handled a result set with more than one row returned.  It didn't properly handle an empty result set. This fixes that.


cmpich (approved):
Looks good.

PR #418, opened 2022-11-14
ZFIN-8268: Can't delete accession from gene edit page

rtaylorzfin:
Return a LinkDisplay for the deleted DBLink (as JSON).


cmpich (approved):
Looks good.

PR #421, opened 2022-11-17
ZFIN-8275: Generate a report of RefSeq matches from UniProt Prob7 file

rtaylorzfin:
Use code from biojava project to parse UniProt Prob7 file.
Put records temporarily into DB and use those records
with some SQL logic to cross-reference the db_link table
and generate report.


cmpich (approved):
Looks good.

PR #424, opened 2022-11-18
Remove GOA-GAF job from the schedule (release 1136).

rtaylorzfin:
Same change as PR #378, but for the 1136 branch.  Main is already fine and 1137 is also fine.


cmpich (approved):
Looks good.

PR #429, opened 2022-11-24
ZFIN-8317 and ZFIN-8318: publications page variations

rtaylorzfin:
Introduce navigation based on a NavigationMenu class with subclass PublicationNavigationMenu for publication-view page. Use the same class to handle the logic of when to show various navigation items / sections and their counts. Hide errata if no errata.

Styling changes:
- align numbers to the right
- change the border
- change the width of navigation

Add option for a feature flag to be enabled by default.


cmpich (approved):
Looks good.

PR #432, opened 2022-11-29
Fix issue with wrapping in the title of the left nav

rtaylorzfin:
Example page with this issue: https://zfin.org/ZDB-TGCONSTRCT-210915-2

Patch to ZFIN-8242


cmpich (approved):
Looks good.

PR #450, opened 2022-12-23
ZFIN-8295: Show all antibody registry IDs 

rtaylorzfin:
Show all antibody registry IDs on the view page if there are multiple. On the edit page, allow updates to contain comma delimited list of registry IDs.


source/org/zfin/marker/presentation/AntibodyMarkerBean.java
```diff
@@ -26,56 +27,9 @@ public class AntibodyMarkerBean extends MarkerBean {
     private int numberOfDistinctComposedTerms ;
     private List<MarkerRelationshipPresentation> antigenGenes;
     private String abRegistryID;
```
cmpich, line 29: Can we remove this accessor  then as it is replaced by the List<String> method?

rtaylorzfin: I deleted abRegistryID as suggested.  It was referenced in antibodyMarkerData.tag, but that file is not used anywhere so I deleted it.

cmpich (approved):
Apart from the one comment / suggestion it looks goof.

PR #452, opened 2022-12-28
ZFIN-8349: Fix html entity for '@' in person table 

rtaylorzfin:
Fix html entity for '@' in person table (57 instances).


cmpich (approved):
Looks good.

PR #454, opened 2022-12-29
ZFIN-8396: Bcrypt hash for passwords

rtaylorzfin:
Allow using md5 if the feature flag is toggled off. This will automatically re-encode the password with bcrypt upon login.  Once feature flag is toggled on, it will no longer allow md5sum authentication and direct people to reset their password.


rtaylorzfin:
closing and will open a PR from a different branch.

PR #455, opened 2022-12-29
Release 1138 Patch: remove unused method "getAllActiveSource"

rtaylorzfin:
We do not need the getAllActiveSource method. Also, the project doesn't compile currently since the method is not implemented in the Impl class--so we must remove it.


cmpich (approved):
Looks good.

PR #457, opened 2022-12-29
Main Branch: ZFIN-8380: Include alias management in antibody edit page

rtaylorzfin:


source/org/zfin/gwt/root/dto/RelatedEntityDTO.java
```diff
@@ -14,8 +14,11 @@
 public class RelatedEntityDTO implements IsSerializable, HasLink, Comparable , Serializable {
 
     // data
+    @JsonView(View.AntibodyAliasAPI.class)
```
cmpich, line 17: DO we need a new Json view for this? Or could we use one of the more generic ones, like the API one?

PR #458, opened 2022-12-30
Release 1138: show all figures

rtaylorzfin:
"Show all figures" link was removed from react component and needs to be included in jsp.


cmpich (approved):
Looks good.

PR #460, opened 2023-01-03
ZFIN-8334 ZFIN-8381 ZFIN-8336 ZFIN-8333 add filtering options for in-situ probes, antibody, phenotype table on term page

cmpich:


rtaylorzfin (approved):
Looks good

PR #461, opened 2023-01-05
ZFIN-8380: Default Publications on PublicationInputs

rtaylorzfin:
Add some default publications to the PublicationInputs on antibody edit page.


cmpich (approved):
Looks good.

PR #467, opened 2023-01-11
Possible fix for git issue in docker build

rtaylorzfin:
Not sure if this is the best fix, but ant tasks were giving the following error
 until I added the "safe" directory configuration here.

```
fatal: detected dubious ownership in repository at '/opt/zfin/source_roots/zfin.org'
To add an exception for this directory, call:
...
```


zfinryanm (approved):
Looks good to me. If the fix works, it seems like a good solution.

PR #470, opened 2023-01-11
ZFIN-8396: Bcrypt hash for passwords

rtaylorzfin:
Use bcrypt for password hashing. Blank out md5 password hashes. Give an "expired password" message when attempting to login on an account with blank password hash.


rtaylorzfin:
Hi Christian,

That’s just the rule of thumb that I’ve come across. I think the idea is that if an attacker got their hands on the DB, they would need to take about 1 second to try every permutation.  This also leaves room for attackers with better hardware.  I don’t think it’s a hard and fast rule. This post recommends ¼ of a second:

https://security.stackexchange.com/a/17238



--
Ryan Taylor (he/him)
Software Developer
The Zebrafish Information Network (ZFIN)
University of Oregon


From: Christian Pich ***@***.***>
Date: Wednesday, January 18, 2023 at 2:16 AM
To: ZFIN/zfin ***@***.***>
Cc: Ryan Taylor ***@***.***>, Author ***@***.***>
Subject: Re: [ZFIN/zfin] ZFIN-8396: Bcrypt hash for passwords (PR #470)

@cmpich approved this pull request.

Looks good.
Why should the server take about 1 second?

PR #472, opened 2023-01-11
ZFIN-8413: getPerson should return null if given null ID

rtaylorzfin:
getPerson should return null if given null for ID


cmpich (approved):
Looks good.

PR #473, opened 2023-01-11
Release 1138 Patch: ZFIN-8413

rtaylorzfin:
getPerson should return null if given null for ID.

This was causing an exception in recording correspondences:

http://github.com/zfin/zfin/blob/ce8d5d9d8f4a63135c79ff7f7599215955cea8a4/source/org/zfin/publication/repository/HibernatePublicationRepository.java#L2599-L2599


zfinryanm (approved):
Looks simple enough to me. Approved.

PR #477, opened 2023-01-12
 Release-1139 Hotfix for ZFIN-8413: getPerson 

rtaylorzfin:
getPerson should return null if given null for ID


cmpich (approved):
Looks good.

PR #479, opened 2023-01-13
ZFIN-8415 paginate and filter over RNA expression table on fish page

cmpich:


source/org/zfin/fish/presentation/FishDetailController.java
```diff
@@ -77,22 +77,7 @@ protected String showFish(@PathVariable String zdbID, Model model) {
         model.addAttribute("phenotypeDisplays", PhenotypeService.getPhenotypeDisplays(phenotypeStatements, "condition", "phenotypeStatement"));
 
         // disease model
```
rtaylorzfin, line 79: I suppose we could get rid of these comments as well

rtaylorzfin (approved):
Looks good!

source/org/zfin/fish/presentation/FishDetailController.java
cmpich: Yes.

PR #482, opened 2023-01-17
Main Branch Hotfix: ZFIN-8421: fix reversion to getOrganizationByZdbID(null)

rtaylorzfin:


zfinryanm (approved):
Look good to me. Same as the previous fix.

PR #483, opened 2023-01-17
ZFIN-8397: Fix rejected email messages due to multiple "To" headers

rtaylorzfin:
Instead of sending a single email with multiple "To" headers, loop over the recipients and send individual emails.

Make it easier to test from the command line.

Example:

```
$ ant writeClassPath
...
 [echo] extended.classpath is written to /tmp/extended.classpath.1620.txt
 [echo] unittest.classpath is written to /tmp/unittest.classpath.1388.txt
...
$ java -cp `cat /tmp/extended.classpath.1620.txt` org.zfin.framework.mail.IntegratedJavaMailSender 'TEST SUBJECT' 'TEST BODY' someusername@zfin.org anotherusername@gmail.com onemoreusername@example.org
```

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -348,6 +348,18 @@
         <path refid="classpath"/>
     </path>
 
+    <target name="writeClassPath">
+        <pathconvert property="classpathProp1" refid="extended.classpath"/>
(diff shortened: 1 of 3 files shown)


cmpich (approved):
Looks good.

PR #487, opened 2023-01-18
ZFIN-5800: Add direct attributions when an alias is added to a marker.

rtaylorzfin:
Also added an endpoint for "ZDB-DALIAS-..." as a convenience.


source/org/zfin/marker/service/MarkerService.java
```diff
@@ -862,6 +862,26 @@ public static boolean markerHasAlias(Marker marker, String alias) {
         return false;
     }
 
+
+    public static MarkerAlias createMarkerAlias(Marker marker, String newAlias, List<String> publicationIDs) {
+        Iterator<String> pubIterator = publicationIDs.iterator();
```
cmpich, line 867: I would write in one line:
String pubID = publicationIDs.iterator().next();
Then you don't have deal with the iterator which has not a strong meaning as pubID.

source/org/zfin/marker/service/MarkerService.java
```diff
@@ -862,6 +862,26 @@ public static boolean markerHasAlias(Marker marker, String alias) {
...
+        Publication publication = publicationRepository.getPublication(pubIterator.next());
+
+        MarkerAlias alias = markerRepository.addMarkerAlias(marker, newAlias, publication);
+
+        // add direct attribution to aliased marker and provided publication
+        infrastructureRepository.insertStandardPubAttribution(marker.getZdbID(), publication);
+
+        while (pubIterator.hasNext()) {
```
cmpich, line 875: Why are you not including the first element's three method calls in the loop?

rtaylorzfin: If I understand correctly, you are asking why the first element is handled with:
```
        MarkerAlias alias = markerRepository.addMarkerAlias(marker, newAlias, publication);
        infrastructureRepository.insertStandardPubAttribution(marker.getZdbID(), publication);
```

while the rest of the elements are handled in the loop with this:
```
        markerRepository.addDataAliasAttribution(alias, publication, marker);
        infrastructureRepository.insertStandardPubAttribution(marker.getZdbID(), publication);
```

This logic was preserved from the previous version of the code.  The "addMarkerAlias" only needs to be called once and that method creates an alias attribution internally so we don't need to call "addDataAliasAttribution" on the first element.  I suppose it doesn't hurt to call "addDataAliasAttribution" twice since it has a check for duplicates.  I'll send another commit for consideration.

source/org/zfin/marker/service/MarkerService.java
```diff
@@ -862,6 +862,28 @@ public static boolean markerHasAlias(Marker marker, String alias) {
...
+            Publication publication = publicationRepository.getPublication(pubID);
+
+            // alias is created with the first reference. others added after the alias is created.
+            if (loopIndex == 0) {
+                alias = markerRepository.addMarkerAlias(marker, newAlias, publication);
+            }
+
+            // add direct attribution to aliased marker and provided publications
+            infrastructureRepository.insertStandardPubAttribution(marker.getZdbID(), publication);
+            markerRepository.addDataAliasAttribution(alias, publication, marker);
```
rtaylorzfin, line 879: Put this in the else of the "if (loopIndex == 0)" to make clear that it is not necessary for the first element.

cmpich (approved):
Looks good.

PR #488, opened 2023-01-19
ZFIN-8425 ZFIN-23 ZFIN-8371 ZFIN-8362 ZFIN-8366 add pagination and filtering to genotype page, etc

cmpich:


home/javascript/react/components/FigureProteinSummary.js
```diff
@@ -5,7 +5,7 @@ import {publicationType} from '../utils/types';
 import PublicationCitationLink from './PublicationCitationLink';
 import CommaSeparatedList from './CommaSeparatedList';
 
-const FigureSummary = ({statistics, fishID}) => {
+const FigureSummary = ({statistics, fishID, markerID =('')}) => {
```
rtaylorzfin, line 8: I don't think the parentheses do anything here: markerID =('') ?

home/javascript/react/components/FigureProteinSummary.js
```diff
@@ -14,19 +14,19 @@ const FigureSummary = ({statistics, fishID}) => {
...
 
-    if (statistics.numberOfFigures < 10) {
+    if (statistics.numberOfFigures < 5) {
         return <CommaSeparatedList>
             {statistics.publications.map(publication => {
                 return <>{
                     publication.figures.map(figure => {
                         return <CommaSeparatedList key={figure.zdbID}>
-                            <><a href={`/${figure.zdbID}`}>{figure.label}</a>{figure.imgless !== false && <CameraIcon/>}</>
+                            <><a href={`/${figure.zdbID}`}>{figure.label}</a>{figure.imgless !== true && <CameraIcon/>}</>
```
rtaylorzfin, line 29: why is the logic reversed from previous version? was this meant to be "=== true" ?

rtaylorzfin (approved):
I only saw one potential issue with the false being swapped to true in the javascript logic. You can refer to my comment there.  Might not be a problem.  Otherwise, it looks great!

home/javascript/react/components/FigureProteinSummary.js
cmpich: Because it was wrong. When the figure is not imageLess then we want to show the icon. double negative here. We probably want to change 'imgless' to 'hasImage', that would be more direct.

home/javascript/react/components/FigureProteinSummary.js
cmpich: Just setting the markerID to an empty string if not set at initialization time (opposed to a null).

rtaylorzfin:
That makes sense

--
Ryan Taylor (he/him)
Software Developer
The Zebrafish Information Network (ZFIN)
University of Oregon


From: Christian Pich ***@***.***>
Date: Tuesday, January 24, 2023 at 11:00 AM
To: ZFIN/zfin ***@***.***>
Cc: Ryan Taylor ***@***.***>, Comment ***@***.***>
Subject: Re: [ZFIN/zfin] ZFIN-8425 add pagination and filtering to genotype page (PR #488)

@cmpich commented on this pull request.

________________________________

In home/javascript/react/components/FigureProteinSummary.js<https://github.com/ZFIN/zfin/pull/488#discussion_r1085773246>:

>          return <CommaSeparatedList>

             {statistics.publications.map(publication => {

                 return <>{

                     publication.figures.map(figure => {

                         return <CommaSeparatedList key={figure.zdbID}>

-                            <><a href={`/${figure.zdbID}`}>{figure.label}</a>{figure.imgless !== false && <CameraIcon/>}</>

+                            <><a href={`/${figure.zdbID}`}>{figure.label}</a>{figure.imgless !== true && <CameraIcon/>}</>

Because it was wrong. When the figure is not imageLess then we want to show the icon. double negative here. We probably want to change 'imgless' to 'hasImage', that would be more direct.

PR #489, opened 2023-01-19
ZFIN-8369: Remove gray background from all figure view.

rtaylorzfin:
Remove gray background from all figure view.


cmpich (approved):
Looks good.

PR #491, opened 2023-01-20
ZFIN-8372: Show images side-by-side for figures with multiple images

rtaylorzfin:
For figures that have multiple images, show them in rows instead of one at a time vertically.

I preserved the behavior of showMultipleMediumSizedImages, though I'm not sure why we want that flag.

On the image page, clicking on the image just redirects you back to the same page.  I changed it to open just the raw image as I think that is the expected behavior.


cmpich (approved):
Looks good.

PR #492, opened 2023-01-21
Postgres 14 compatibility tweak

rtaylorzfin:
My local pg environment did not like the variable to be named "overlaps" due to conflict with keyword.


cmpich (approved):
Looks good.

PR #495, opened 2023-01-23
Security Context Fix for Testing -- version 1

rtaylorzfin:
This seems like the right approach to getting failing tests to work due to autowiring the bcrypt bean failing.

We need to tell the test to load the security.xml file.  The problem is the security.xml file in the source tree contains placeholders that get replaced later (${SECUREPORT}, etc.).  That makes this approach a bigger restructuring.


rtaylorzfin:
Closing for now. We could revisit this later.

PR #496, opened 2023-01-23
Security Context Fix for Testing -- version 2

rtaylorzfin:
Instead of autowiring the bcrypt password encoder in profile service, just use a constructor to instantiate it.


cmpich (approved):
Looks good.

PR #498, opened 2023-01-25
ZFIN-8333 fix up feature_stats table with antigen genes

cmpich:


source/org/zfin/db/postGmakePostloaddb/1139/ZFIN-8333.sql
```diff
@@ -0,0 +1,32 @@
+--liquibase formatted sql
+--changeset cmpich:ZFIN-8333
+
+create temp table antibody_temp
```
rtaylorzfin, line 4: It's a matter of personal preference, I suppose, but you can combine the create temp table and insert with:

```
select mrel_mrkr_2_zdb_id, mrel_mrkr_1_zdb_id 
INTO TEMP TABLE antibody_temp
from marker_relationship
where mrel_mrkr_2_zdb_id ~'ATB';
```

rtaylorzfin (approved):
Looks good

source/org/zfin/db/postGmakePostloaddb/1139/ZFIN-8333.sql
cmpich: I like it!

PR #506, opened 2023-01-31
ZFIN-8442: Fix the dialog styling issues

rtaylorzfin:
Fix the dialog styling issues as a result of the upgrade from jquery-modal from 0.5.9 to 0.9.2.

See: https://github.com/kylefox/jquery-modal/blob/master/jquery.modal.css


cmpich (approved):
I cannot validate this... Probably ok.

PR #507, opened 2023-02-01
ZFIN-8392: Remove duplicates from MarkerEditDbLinks dialog

rtaylorzfin:
Add the fdb_db_name column to the serialization of ForeignDB object. It already pulled that column, but transformed it to an enum.  This adds the column again, but as a string.

Remove duplicate entries from the dialog for MarkerEditDbLink (of identical options, favor the earliest zdbID).

Clean up the DisplayGroupRepository for modern hibernate.

lombokify where appropriate

### DIFF
--- a/home/javascript/react/containers/MarkerEditDbLinks.js
+++ b/home/javascript/react/containers/MarkerEditDbLinks.js
@@ -57,6 +57,40 @@ const MarkerEditDbLinks = ({markerId, group = 'other marker pages'}) => {
         return null;
     }
 
+    function groupBy(items, getKey) {
+        return items.reduce((result, item) => {
+            const key = getKey(item);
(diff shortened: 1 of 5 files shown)


home/javascript/react/containers/MarkerEditDbLinks.js
```diff
@@ -57,6 +57,40 @@ const MarkerEditDbLinks = ({markerId, group = 'other marker pages'}) => {
         return null;
     }
 
+    function groupBy(items, getKey) {
```
rtaylorzfin, line 60: Let's move this logic into the API if not the database.

rtaylorzfin:
Closing. The fix in PR #514 is a better solution.

PR #510, opened 2023-02-02
Docker: Removed container name and port collisions to support multipl…

zfinryanm:
…e zfin.org stacks in a single docker instance.


zfinryanm:
Not sure what to do about the Mailhog config.

rtaylorzfin (approved):
Looks good!

PR #518, opened 2023-02-06
ZFIN-8392: Second revision - we no longer need the "other pages" option on STR edit page

rtaylorzfin:
We do not need the ability to add "other pages" to the STR edit page.  See discussion on jira ticket.


cmpich (approved):
Looks good.

PR #521, opened 2023-02-08
ZFIN-8462: load author names fix

rtaylorzfin:
Fix issue with some first names and last names getting scrambled. 
Delete potentially erroneous rows -- to be rebuilt later.


cmpich (approved):
Looks good.

PR #522, opened 2023-02-08
ZFIN-8461: publication page left nav count root-only

rtaylorzfin:
- Only show counts on the publication page left nav if user is root


cmpich (approved):
Looks good.

PR #525, opened 2023-02-10
Release 1139: Hotfix for ZFIN-8464 and ZFIN-8468

rtaylorzfin:
Allow deleting a dblink without attribution from gene page.
Trim whitespace from pub ID when submitted for attribution.


cmpich (approved):
Looks good.

PR #530, opened 2023-02-17
ZFIN-8337: spring minor upgrade

rtaylorzfin:
Remove spring-*-4.2.2*.jar in favor of spring 4.3.5 jars. The exception is the spring-security jars which use a different numbering scheme for some reason. Added some logging to help with eventually deprecating and removing spring-ws (SOAP) jars in the future.

Add some logging to help figure out if these methods are safe to deprecate.


cmpich (approved):
Looks good.

PR #533, opened 2023-02-20
ZFIN-8450: Adjust the height of jbrowse on mapping details page.

rtaylorzfin:
Add a height attribute to jbrowse component so we can set a shorter height on the mapping details page.


cmpich (approved):
Looks good.

PR #536, opened 2023-02-22
ZFIN-8456: Tweaks to formatting on the figure page:

rtaylorzfin:
Change Title to Figure Name eg. “Fig. 1”

Move the current zdb ID to the second line just below title and re-label as “ZDB ID”

Change linking such that from all figure view, clicking a figure image should take you to a figure page, then clicking an image from a figure page should take you to an image page.

On the figure page summary, change “Figures” to “Other Figures” and order them sequentially (with a show all link at the beginning)

Change wording on “All figure page → Show All” to say “Back to All Figure Page” (to avoid confusion with the expanding component that says “all…”)


cmpich (approved):
Looks good.

PR #539, opened 2023-02-23
ZFIN-8482 (release-1140 hotfix): Show captions on figures

rtaylorzfin:
The tags imagesAndCaption.tag and imagesAndCaptionPrototype.tag were reversed (the prototype one was used on legacy pages, and the non-prototype one was used on prototype pages). This reverses that to avoid confusion.

Also, fix bug where captions were not being displayed even though showCaption was true.


zfinryanm (approved):
Seems fine.

zfinryanm:
Christian should probably do the code review for the pull request against main.

PR #547, opened 2023-03-02
ZFIN-8286: Working config for Jenkins under docker.

zfinryanm:


rtaylorzfin (approved):
Looks good to me

server_apps/jenkins/jobs/Check-1-To-Many-ENSDARG-ZFIN/config.xml
```diff
@@ -97,11 +97,5 @@
     </hudson.plugins.emailext.ExtendedEmailPublisher>
   </publishers>
   <buildWrappers>
-    <EnvInjectBuildWrapper plugin="envinject@1.89">
```
cmpich, line 100: I thought we need this plugin, don't we?

zfinryanm: We do but it wasn't working correctly under Jenkins 2.375. Not with docker, at least. I switched it to being injected globally and that worked, so we don't need it set on each individual job.

zfinryanm: See <globalNodeProperties> in the global config.xml.

cmpich (approved):
Looks good to me. Need to test it out once it is in main.

PR #550, opened 2023-03-02
ZFIN-8470: construct edit YFP -> lanYFP

rtaylorzfin:
Change association between constructs and YFP to lanYFP.


cmpich (approved):
Looks good.

PR #555, opened 2023-03-07
ZFIN-8237: Add nav section. Change resolution to radio button.

rtaylorzfin:
I realized during the sprint review demonstration that we were missing the section header in the left navigation. Also, the resolution should be a radio input instead of a checkbox.


cmpich (approved):
Looks good.

PR #559, opened 2023-03-08
ZFIN-8237: Fix to radio button handling

rtaylorzfin:
The onchange handler was copied from the checkbox logic. For radio buttons, it's slightly different. This fixes that issue. Also removed the spinner for resolutions.


cmpich (approved):
Looks good.

PR #560, opened 2023-03-08
ZFIN-8449 ZFIN-8502 add probe table paginated / filterd on publication page, add probe ID to call all-figure page

cmpich:


rtaylorzfin (approved):
Looks good

PR #564, opened 2023-03-13
optimize drop-and-reload process for UI DB tables

cmpich:


cmpich:
I am trying to reduce the time the re-load transaction is causing table / column locking.
Before the retrieval of all records was part of the transaction. Now, retrieval is its own tx and then removal and re-generation another one.

PR #565, opened 2023-03-13
ZFIN-8478 add filtering on phenotype and expression tables on figure page

cmpich:


home/javascript/react/containers/FigureExpressionAntibodyTable.js
```diff
@@ -64,7 +64,7 @@ const FigureExpressionAntibodyTable = ({figureId}) => {
         <DataTable
             columns={columns}
             dataUrl={`/action/api/figure/${figureId}/antibody-labeling`}
-            rowKey={row => row.antibody.zdbID}
+            rowKey={() => Math.random()}
```
rtaylorzfin, line 67: Why use a random rowKey?

home/javascript/react/containers/FigurePhenotypeTable.js
```diff
@@ -63,7 +67,7 @@ const FigurePhenotypeTable = ({url, hideFigureColumn = false, navigationCounter,
         <DataTable
             columns={columns}
             dataUrl={url}
-            rowKey={row => row.rowKey}
+            rowKey={() => Math}
```
rtaylorzfin, line 70: Is this line supposed to include .random()? I'm still not sure the reason why random is better than a unique id

home/javascript/react/containers/FigureExpressionAntibodyTable.js
cmpich: Because the unique row id is a combination of multiple columns.

PR #567, opened 2023-03-13
Gradle Feature Deprecation: "enabled" is deprecated and replaced with "required".

rtaylorzfin:
See: https://docs.gradle.org/7.5.1/javadoc/org/gradle/api/reporting/Report.html

Screenshot 2023-03-13 at 11 40 24 AM

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -974,8 +974,8 @@ test {
     }
 
     reports {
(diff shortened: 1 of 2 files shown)


zfinryanm (approved):
Looks good.

PR #568, opened 2023-03-14
ZFIN-8475: Image Prototype Tweaks

rtaylorzfin:
Hide Figure Caption if there is no caption.

Fix acknowledgments section on image page to match figure page


cmpich (approved):
Looks good.

PR #574, opened 2023-03-17
ZFIN 8474: Gene-Association-File-Export_w remove one artifact

rtaylorzfin:
Remove gp2protein.zfin.gz from artifacts


zfinryanm (approved):
Nice.

PR #580, opened 2023-03-22
ZFIN-8529: Change geneontology.org URLs to match new format.

rtaylorzfin:
Change URLs to match new format.


zfinryanm (approved):
Looks good to me.

PR #586, opened 2023-03-28
ZFIN-8537: Human Diseases Section Failing on Publication Page

rtaylorzfin:
Fixed by checking for undefined or null before map.
However, I noticed that the PublicationDiseaseTable react component (https://github.com/zfin/zfin/blob/main/home/javascript/react/containers/PublicationDiseaseTable.js#L25-L25)
 is expecting an "evidenceCodeList" that is undefined. Should we be providing that list from the backend?

Here's the backend URL:
https://zfin.org/action/api/publication/ZDB-PUB-181215-18/diseases


zfinryanm (approved):
Looks good.

PR #589, opened 2023-04-04
ZFIN-8499: Refactorings of UniProt and RefSeq Matching

rtaylorzfin:
Add the REFSEQ match from sp_check for sp_parser.
Refactor pre_loadsp.pl for readability.
Add bugfix for weird issue with SO:0001217 (see comments on ZFIN-8499) Wrap primary logic in main method.
Refactor to not use temporary files when we can just use buffer variables.
Fix match to end-of-record: the regex "/\/\//" was erroneously matching the occasional URL lines, eg: CC       URL="https://web.expasy.org/spotlight/back_issues/210/";
refactor embl matching if clause to method.
move chunks of logic into subroutines
Add some checksum info for versioning
Add better logging to problem files. refseq problem files are now prob9 and prob10.
Clean up output of REFSEQ match.
Improve performance of db lookups with cache.


cmpich (approved):
Looks good.

PR #596, opened 2023-04-10
Fix solr start/stop/restart targets. Similar to ZFIN-8511.

rtaylorzfin:


zfinryanm (approved):
Looks good. Thanks for catching it.

PR #599, opened 2023-04-11
Switch from node-sass to sass

rtaylorzfin:
We've talked about switching to this version of sass before. I think we were waiting to make sure the other npm packages we upgraded don't cause problems. Now might be a good time to switch.  It would be nice to have the pure js implementation so we don't depend on python, and C for npm installation steps.


cmpich (approved):
Looks good.

PR #602, opened 2023-04-13
ZFIN-8449 remove rating column and add clone type column, support supplemental data support for filter selections

cmpich:


rtaylorzfin (approved):
Looks good

PR #615, opened 2023-04-17
ZFIN-8540: Fix the word wrapping of disease terms 

rtaylorzfin:
Fix the word wrapping of disease terms in the top left corner. Hide navigation item for alliance data if flag is disabled.


cmpich (approved):
Looks good.

PR #616, opened 2023-04-18
ZFIN-8373 ZFIN-8504 ZFIN-8515 suppress 'show all Figure' link when more than 50 figures per publication, fix order of section, enforce EaP rules

cmpich:


rtaylorzfin (approved):
Looks good

PR #626, opened 2023-04-24
ZFIN-8589 ZFIN-8591 add ontology search box to term page, add citations section to DO term page

cmpich:


rtaylorzfin (approved):
Looks good

PR #628, opened 2023-04-25
ZFIN-8586: load-flank-seq job fix

rtaylorzfin:
- Add some more output to the Flank Sequence Processor job
- When transactions are committed mid-job, open another transaction
- Cleanup: unused imports and lombokify


cmpich (approved):
Looks good.

PR #631, opened 2023-04-26
ZFIN-8517: Update to the NCBI One Way Matches report

rtaylorzfin:
Ken requested the related RNA sequences for this report. We have somewhat complex logic for retrieving RNA sequences per marker from the gene page.  Instead of re-inventing that logic through SQL queries, I use the existing logic and iterate over each row to get the related RNA sequences.  It's slow, but it works.


cmpich (approved):
Looks good.

PR #632, opened 2023-04-26
ZFIN-8575: Hide figure label on all-figure page as it is redundant.

rtaylorzfin:


rtaylorzfin:
@cmpich do you know why there is this comment on this ticket:

"Can be done only when we move this page to prod."

cmpich (approved):
Looks good.

PR #634, opened 2023-04-27
ZFIN-8579: Hide acknowledgments and orientation as appropriate on image pages.

rtaylorzfin:


source/org/zfin/framework/presentation/ImageNavigationMenu.java
```diff
@@ -63,11 +64,18 @@ public void setModel(Model model) {
         this.setHidden(NavigationMenuOptions.COMMENTS, StringUtils.isEmpty(image.getComments()));
         this.setHidden(NavigationMenuOptions.DEVELOPMENTAL_STAGE, image.getImageStage() == null || image.getImageStage().getStart() == null);
         this.setHidden(NavigationMenuOptions.ORIENTATION,
-        "not specified".equals(image.getPreparation()) &&
-            "not specified".equals(image.getForm()) &&
-            "not specified".equals(image.getDirection()) &&
+        "not specified".equals(image.getPreparation()) ||
```
cmpich, line 67: I wonder if there is a better way for these conditions. This way it is 'hard' to read.

rtaylorzfin: Good point! Not only is it hard to read, but upon reviewing, I believe it is wrong!

I was preserving the logic from the jsp that has this:
```
        <c:if test="${image.preparation ne 'not specified' && image.form ne 'not specified' && image.direction ne 'not specified' && image.view ne 'not specified'}">
```

So the jsp logic only shows the image orientation if none of the orientation fields have 'not specified'.  But it seems to me that it should show the orientation if any of those fields contain information.  I fixed that logic and made a change to make it more readable.

To illustrate the issue with the logic, this image (ZDB-IMAGE-020802-27) is currently hiding the orientation information on PROD:

Screenshot 2023-04-28 at 10 14 38 AM


And with the logic change (on prototype page):

Screenshot 2023-04-28 at 10 14 02 AM

PR #637, opened 2023-04-29
Ran the inspection "Pattern variable can be used" against our codebase.

rtaylorzfin:


cmpich (approved):
Nice! I did not know you can assign the typed version in the instanceOf check, called pattern match.

rtaylorzfin:
> Nice! I did not know you can assign the typed version in the instanceOf check, called pattern match.

This was new to me too! Pretty cool

PR #648, opened 2023-05-08
allow for executing individual indexers through the dashboard

cmpich:


source/org/zfin/indexer/presentation/IndexerDashboardController.java
```diff
@@ -13,4 +17,21 @@ protected String showIndexerDashboard(Model model) {
         return "indexer/indexer-view";
     }
 
+    @RequestMapping(value = "/runIndexer/{indexerName}", method = RequestMethod.GET)
```
rtaylorzfin, line 20: Should this be a POST method? I've had issues in the past where my browser tries to pre-fetch pages with GET requests.

cmpich: Yeah, I agree. Do you know how to turn the hyperlink easily into a POST request?

rtaylorzfin: I created a quick PR that should do the trick:

https://github.com/ZFIN/zfin/pull/657/files

Now that I'm looking at it again, we could probably clean that up a little bit more by omitting the href value.

PR #656, opened 2023-05-22
Fix for running "ant tomcat-restart" from within compile container.

rtaylorzfin:
"ant tomcat-stop" and "ant tomcat-start" stopped working from within the docker container for me. With docker versions 23 and 24, the stop command would fail silently. With earlier docker versions, I got the following error message:

```
zfin.org:.../zfin.org/docker$ docker compose up tomcat
invalid project name "zfin.org": must consist only of lowercase alphanumeric characters, hyphens, and underscores as well as start with a letter or number
```

Also, added tabs to tomcat.xml.


zfinryanm (commented):
Sad to see the change. Would it be better to use an underscore, i.e. zfin_org? I'm fine with either.

zfinryanm (approved):
Looks good.

PR #661, opened 2023-05-25
ZFIN-8620: Clean up marker relationships

rtaylorzfin:
This liquibase script removes the unnecessary multiple attributions for marker relationships for constructs.


cmpich (approved):
I assume all these IDs are correct.

rtaylorzfin:
I am keeping the oldest ID per MREL in the record_attribution table. I confirmed with Holly that this is the way to go. I also did a simple cross-check to make sure that the oldest recattrib ID also corresponds to the oldest ZDB-PUB ID.

PR #662, opened 2023-05-25
ZFIN-8598 and ZFIN-8616: Add validation check disallowing a deficiency with "is allele of" relationship

rtaylorzfin:
Also disallow more than one "is allele of" for any mutant type.


cmpich (approved):
Looks Good.

PR #678, opened 2023-06-05
ZFIN-8676: Fix issue from switching from setString to setParameter on hql queries.

rtaylorzfin:
For example, we had to change queries like this:

```
hql = "...where m.markerType = :markerTypeName";
...
query.setString("markerTypeName", Marker.Type.MRPHLNO.name());
```

to

```
hql = "...where m.markerType.name = :markerTypeName";
...
query.setParameter("markerTypeName", Marker.Type.MRPHLNO.name());
```


cmpich (approved):
Looks good.

PR #682, opened 2023-06-08
ZFIN-8640

cmpich:
I also added the ShowDevInfo component to display info about the endpoin (for ROOT only users)


rtaylorzfin (approved):
Looks good

PR #684, opened 2023-06-08
ZFIN-8623: Fix image thumbnails

rtaylorzfin:
Add a gradle task that fixes missing image thumbnails by converting full-size images to thumbnails and medium images. It retrieves a list of all images associated with figures, checks for missing thumbnails and mediums, and performs the necessary imagemagick command if required.


cmpich (approved):
Looks good.

PR #697, opened 2023-06-15
ZFIN-8492: self binding gene

rtaylorzfin:
Add sp_load step to include the gene's own ID as an inference group member if self-binding.


cmpich (approved):
Looks good.

PR #702, opened 2023-06-20
ZFIN-8684 rename genotype according to DB genotype naming function

cmpich:


rtaylorzfin (approved):
Looks good. Thanks for grabbing this one!

PR #704, opened 2023-06-20
ZFIN-8573 - Main Branch - Bugfix on validation: call "getFeatureAssembly" instead of "getAssembly"

rtaylorzfin:


cmpich (approved):
Looks good.

PR #705, opened 2023-06-20
ZFIN-8573 - Release 1144 - Bugfix on validation: call "getFeatureAssembly" instead of "getAssembly"

rtaylorzfin:


zfinryanm (approved):
Looks good.

PR #707, opened 2023-06-23
ZFIN-8491: Change located_in to part_of 

rtaylorzfin:
Change located_in to part_of if relationship is for a descendant of "protein containing complex".


cmpich (approved):
Looks good.

PR #710, opened 2023-06-27
ZFIN-8642 fix pagination issue: scrollable.getRowCount() == 0 instead of -1 before first record

cmpich:


rtaylorzfin (approved):
Looks good

PR #711, opened 2023-06-27
ZFIN-8625: NCBI patch: Fix reference to undefined value

rtaylorzfin:
On trunk, was generating an error due to an undefined value in $GenPeptsToLoad{$GenPept} (ERROR - GenPept ABC43274 not in GenPeptsToLoad hash)


cmpich (approved):
Looks good.

PR #716, opened 2023-06-29
ZFIN-8673 fix panther load script

cmpich:


rtaylorzfin (approved):
Looks good. Is the sql case sensitive?

PR #717, opened 2023-06-29
ZFIN-8523 move bin3 pubs into tox natural products bin

cmpich:


rtaylorzfin (approved):
Looks good. FYI, you can use "ILIKE" for case insensitive "LIKE".

PR #719, opened 2023-06-29
Backport of UniProt Changes from release-1145 to release-1144

rtaylorzfin:
These are changes that were made in release-1145 that relate to the UniProt load. This patch is to bring those changes into 1144 for running on PROD before deploying release-1145.


cmpich (approved):
Looks good.

PR #723, opened 2023-07-03
ZFIN-8444: Clean up foreign_db_contains_display_group_member and foreign_db_contains

rtaylorzfin:


source/org/zfin/db/postGmakePostloaddb/1146/ZFIN-8444.sql
```diff
@@ -0,0 +1,13 @@
+--liquibase formatted sql
+--changeset rtaylor:ZFIN-8444.sql
+
+DELETE FROM foreign_db_contains_display_group_member WHERE fdbcdgm_fdbcont_zdb_id in ('ZDB-FDBCONT-041215-1', 'ZDB-FDBCONT-090929-8', 'ZDB-FDBCONT-220124-1', 'ZDB-FDBCONT-130823-1', 'ZDB-FDBCONT-140109-1') ;
```
cmpich, line 4: Why do you have three more foreign db contains in the inclause? I thought just the two you are deleting in the next command.

rtaylorzfin: Those are the ones that we found when we were exploring this in CS meeting.  When we tried to add a foreign key to the foreign_db_contains_display_group_member we got an error because those values do not correspond to any values in foreign_db_contains.

PR #728, opened 2023-07-11
ZFIN-7002: convert gbrowse links to jbrowse

rtaylorzfin:
There are a few places that generate links to a gbrowse interface.
The links are of the form "/gbrowse/Zv9/?name=ZDB-GENE-..." or "https://trunk.zfin.org/action/gbrowse/?name=ENSDARG00..."
This commit adds the logic so that if jbrowse flag is enabled, the generated links are of the form "/action/jbrowse/byName?name=...".
The controller that handles that endpoint determines the parameters needed for a jBrowse link and it redirects to the appropriate
jBrowse interface.


source/org/zfin/search/service/RelatedDataService.java
```diff
@@ -207,6 +207,15 @@ public List<String> getRelatedDataLinks(SearchResult result) {
         return links;
     }
 
+    private String makeGenomeBrowserLinkByID(String id) {
+        String url = GenomeLocation.Source.ZFIN_Zv9.getUrl() + id;
```
cmpich, line 211: Is it always Zv9 assembly?

rtaylorzfin: Good question. In this case, I'm preserving the original logic. In the original version, it was only Zv9. So as long as the original version is correct, this one should be. I could make it more explicit by renaming the method to "makeZv9GenomeBrowserLinkByID".

You can see the original version on lines 100 and 111:
```
gBrowseLink = makeLink(GENOME_BROWSER, "/" + ZfinPropertiesEnum.GBROWSE_ZV9_PATH_FROM_ROOT + "?name=" + id);
```

PR #733, opened 2023-07-13
Release 1145: ZFIN-8726: Fix record_attributions to deleted (merged) publications

rtaylorzfin:


cmpich (approved):
Looks good. Though, we should put the function into a sql file into the source tree as we do for all functions and views.

PR #736, opened 2023-07-14
Image Page Cleanup

rtaylorzfin:
I came across some code that can be cleaned up.

The jsp "test" code is unnecessary because section.tag checks for a navigationMenu and uses its logic for whether to display a section.


cmpich (approved):
Looks good.

PR #738, opened 2023-07-14
ZFIN-8736: Change Genotype-Name-Function-Mismatch to weekly

rtaylorzfin:
Change Genotype-Name-Function-Mismatch to weekly.

Simplify the definition of the database view "genotype_name_components_from_relationships" that is used by the function "get_genotype_display" to determine genotype names.

Also, added feature_zdb_id, mrkr_zdb_id to the same view for use in debugging.


cmpich (approved):
Looks good.

PR #741, opened 2023-07-19
ZFIN-8731 and ZFIN-8728: Change from GENE to NCRNAG

rtaylorzfin:


source/org/zfin/db/postGmakePostloaddb/1146/ZFIN-8728.sql
```diff
@@ -0,0 +1,67 @@
+--liquibase formatted sql
+--changeset rtaylor:ZFIN-8728.sql
+
+INSERT INTO zdb_active_data (zactvd_zdb_id) VALUES ('ZDB-NCRNAG-030131-2');
```
cmpich, line 4: Why do you use the index of 2?
ZDB-NCRNAG-030131-1
should be available for that day...

rtaylorzfin: See ZFIN-8731.sql

cmpich (approved):
Looks good.

PR #747, opened 2023-07-26
ZFIN-8524 ZFIN-8752 move bin3 tox papers to Drug category, make tmp tables TEMP tables

cmpich:


source/org/zfin/db/postGmakePostloaddb/1147/ZFIN-8524.sql
```diff
@@ -0,0 +1,155 @@
...
+    cur_zdb_id text
+);
+
+-- match on title
+select count(*)
+from pub_tracking_history,
+     publication
+where zdb_id = pth_pub_zdb_id
+  and pth_location_id = 6
+  and (lower(title) like '%therap%' or lower(title) like '%preclincial drug%' or lower(title) like '%efficacy%')
```
rtaylorzfin, line 17: Should "preclincial" be "preclinical"?

cmpich: Good catch!

PR #755, opened 2023-08-02
ZFIN-8701: getPDFandImages.groovy improvement

rtaylorzfin:
Add optional positional arguments to getPDFandImages.groovy so that you can tell it which publications to download images for. Example:

./getPDFandImages.groovy ZDB-PUB-201231-1 ZDB-PUB-201231-2


cmpich (approved):
Looks good.

PR #765, opened 2023-08-16
ZFIN-8073, ZFIN-8723: UniProt Diff Tool & UniProt Backfill Record Attributions

rtaylorzfin:
This adds a task (uniprotReleaseDiffTask) and jenkins job (UniProt-Diff-Report) that compares two UniProt releases and generates a report that shows the differences (only including changes we care about at ZFIN). 

We also backfill some record attributions needed as outlined in ZFIN-8723


rtaylorzfin:
Closing so that we can separate into 2 PRs

PR #766, opened 2023-08-17
ZFIN-8745 and ZFIN-8746: Change Gene Type to NCRNAG

rtaylorzfin:
Change Genes to NCRNAG:

ZDB-GENE-030131-3332
ZDB-GENE-090313-169


cmpich (approved):
Nice work. This way we can easily convert future genes to ncrnas.

PR #768, opened 2023-08-18
ZFIN-8722 patch: COLLATE issue

rtaylorzfin:
on trunk, postgresql did not like using column numbers in the order by clause in combination with "COLLATE".
on cell, this did work.


cmpich (approved):
Looks fine.

PR #769, opened 2023-08-18
ZFIN-8723: UniProt Load - Add record attribution to db_links even if they exist

rtaylorzfin:
Add record attribution to db_links if they exist and match what would've been new records coming in. Add manual inserts to compensate for the missing records from the last UniProt load.


source/org/zfin/db/postGmakePostloaddb/1147/ZFIN-8723.sql
```diff
@@ -0,0 +1,70 @@
...
+-- These dblinks have been reviewed on ticket ZFIN-8723 and are ready to be added to the database.
+-- They correspond to records that were previously considered redundant because the UniProt accessions
+-- were already manually added. However, we want to preserve the information that shows these records
+-- are also matched to UniProt with the automated process.
+
+CREATE TEMP TABLE new_recattrib (
+    "dblink_zdb_id" text
+);
+
+INSERT INTO new_recattrib VALUES
```
cmpich, line 13: It would be nice know where those came from. Could you write a query that select theses DB_LINK records? If no east query can be written then this is ok.

cmpich (approved):
Looks good.

PR #770, opened 2023-08-18
ZFIN-8073: UniProt Diff Tool 

rtaylorzfin:
This adds a task (uniprotReleaseDiffTask) and jenkins job (UniProt-Diff-Report) that compares two UniProt releases and generates a report that shows the differences (only including changes we care about at ZFIN).


source/org/zfin/uniprot/UniProtReleaseDiffTask.java
```diff
@@ -0,0 +1,190 @@
...
+
+
+    public void runTask() throws IOException, BioException, SQLException {
+        initAll();
+        initWorkspace();
+
+        //handle input file set 1
+        //filter the input files using a filter object (FilterTask) and concatenate to create a single filtered file
+        Path filteredFile1 = combineAndFilterInputFileSet(inputFilenameSet1, "filtered_set1.dat");
+        System.out.println("Filtered file 1: " + filteredFile1.toAbsolutePath());
```
cmpich, line 79: Could you use log4j2 here instead of System .out statements?

cmpich (approved):
Looks great with the System.out changes.

PR #775, opened 2023-08-23
ZFIN-8720: fix superscript issue in safari

rtaylorzfin:
Fix for issue with safari (and only some versions of safari) occasionally superscripting (or appearing to superscript) the last element of a comma separated list.


cmpich (approved):
Looks good.

PR #777, opened 2023-08-25
ZFIN-8351: deadlock updates

rtaylorzfin:
Since the deadlock always occurs upon trying to delete genotype_figure_fast_search, this code avoids dropping the table and instead renames it and its indexes.

The renamed version of the table is later be deleted by regenPhenotypeMartCleanup.sh


cmpich (approved):
Looks good.

PR #785, opened 2023-08-31
ZFIN-8351: 2 More Tables to Rename Instead of Drop

rtaylorzfin:
Apply the same logic for mutant_fast_search and clean_expression_fast_search tables as we did for genotype_figure_fast_search by renaming instead of dropping.


lib/DB_functions/Regen_genox/regen_genox_finish_marker.sql
```diff
@@ -46,6 +46,26 @@ begin
 
      -- let errorHint = "drop mutant_fast_search table ";
       drop table mutant_fast_search;
```
rtaylorzfin, line 48: remove this line

PR #794, opened 2023-09-06
Release 1147 - ZFIN-8351: deadlock updates (combined commits)

rtaylorzfin:
Since the deadlock always occurs upon trying to delete one of:

- genotype_figure_fast_search
- mutant_fast_search
- clean_expression_fast_search (actually, haven't seen this one cause deadlock yet)

this code avoids dropping the tables and instead renames them and their indexes. The renamed version of the table/indexes is later be deleted by regenPhenotypeMartCleanup.sh


cmpich (approved):
Looks good.

PR #797, opened 2023-09-07
ZFIN-8351: One more bugfix:

rtaylorzfin:
- forgot to include the "select" part of running a function
- take into account constraints when dropping indexes


cmpich (approved):
Looks good.

PR #798, opened 2023-09-07
ZFIN-7053: Make external links open in new tab on gene page.

rtaylorzfin:
- In JSPs, replace '<a href...' tags with '<zfin2:externalLink...' where appropriate
- In React components, replace the '<a ...' jsx tags with a new react component '<ExternalLinkMaybe ...' that opens in a new tab if not a local link


cmpich (approved):
Looks good.

PR #799, opened 2023-09-07
ZFIN-8809: Change ZDB-LINCRNAG-050208-65 to ZDB-GENE-050208-804

rtaylorzfin:
expand convert_gene_to... functionality to support any marker type.

- Also, if expression_experiment2 table references marker, propagate that change

### DIFF
--- a/source/org/zfin/db/postGmakePostloaddb/1148/ZFIN-8809.sql
+++ b/source/org/zfin/db/postGmakePostloaddb/1148/ZFIN-8809.sql
@@ -1,0 +1,4 @@
+--liquibase formatted sql
+--changeset rtaylor:ZFIN-8809.sql
+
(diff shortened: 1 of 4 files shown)


source/org/zfin/db/postGmakePostloaddb/1148/ZFIN-8809.sql
```diff
@@ -0,0 +1,4 @@
+--liquibase formatted sql
+--changeset rtaylor:ZFIN-8809.sql
+
+select convert_gene_to_type('ZDB-LINCRNAG-050208-65','GENE');
```
cmpich, line 4: The name does not describe fully the intention: 
convert gene to type
but you are converting a ncr to a gene. Should it be convert marker to marker or something similar?

cmpich (approved):
Looks good.

PR #802, opened 2023-09-11
ZFIN-8770: no adding ncbi genes manually

rtaylorzfin:


home/javascript/react/containers/MarkerEditDbLinks.js
```diff
@@ -82,9 +82,10 @@ const MarkerEditDbLinks = ({markerId, group = 'other marker pages'}) => {
                         validate={value => value ? false : 'A database is required'}
                     >
                         <option value='' />
-                        {databases.value.map(database => (
-                            <option value={database.zdbID} key={database.zdbID}>{database.name}</option>
-                        ))}
+                        {databases.value
```
rtaylorzfin, line 85: Look at the backend and check if we need to add some logic to the backend.

rtaylorzfin: Better to handle this on the backend

rtaylorzfin:
I tried a different approach with the latest commits.  I added "permissions" columns to the foreign_db_display_group_member table.  Then I marked the "can_add" permission as false for the row that puts NCBI Genes in the "other markers" display group.  This signifies that curators are not allowed to add NCBI Genes.

cmpich (approved):
Looks good.

PR #805, opened 2023-09-12
Revert "ZFIN-8345 (#753)"

rtaylorzfin:
I had to revert this change during release-1147 deploy.

Screenshot 2023-09-12 at 11 43 09 AM


cmpich (commented):
I am going to fix this. Do not revert on main.

PR #812, opened 2023-09-21
ZFIN-8738 convert all-figures-view to /all-figures

cmpich:


server_apps/apache/inc-redirect
```diff
@@ -98,7 +98,7 @@ RewriteCond %{QUERY_STRING}     ^Mival=aa-fxfigureview.apg&OID=([\w-]+)$    [NC]
 RewriteRule ^.*$ /%1? [NC,R=301]
 
 RewriteCond %{QUERY_STRING}     ^Mival=aa-fxallfigures.apg&OID=([\w-]+)$    [NC]
-RewriteRule ^.*$ /action/figure/all-figure-view/%1? [NC,R=301]
+RewriteRule ^.*$ /action/publication/%1/all-figures? [NC,R=301]
```
rtaylorzfin, line 102: should we add a general rewrite rule in case someone follows an old link?

Something like:
/action/figure/all-figure-view/(.*)?(.*) -> /action/publication/$1/all-figures?$2

cmpich: Good idea. However, I'd like to remove the really old redirects from Informix,i.e. Mival=aa- and that sort.

rtaylorzfin (approved):
Looks good!

PR #819, opened 2023-09-22
ZFIN-8824

cmpich:


server_apps/data_transfer/PUBMED/loadNewPubs.sql
```diff
@@ -21,8 +21,8 @@ create temp table tmp_pubs (
   status text,
   pubType text);
 
-\copy tmp_pubs from '<!--|TARGETROOT|-->/server_apps/data_transfer/PUBMED/parsePubs.log';
-\copy (select zdb_id from tmp_pubs,publication where accession_no = pmid) to '<!--|TARGETROOT|-->/server_apps/data_transfer/PUBMED/pubAlreadyinZFIN.txt' delimiter '|';
+\copy tmp_pubs from '/Users/cmpich/projects/zfin/schlapp/server_apps/data_transfer/PUBMED/parsePubs.log';
```
rtaylorzfin, line 24: Looks like you lost '''.  Could we use a relative path?

cmpich: Good catch. I think you can make the path of the file to read relative, not sure if that works for the output file write.

PR #823, opened 2023-09-26
ZFIN-8844 Construct linkml

cmpich:


rtaylorzfin (approved):
This looks good to me, though I don't really know what the linkml is supposed to look like. I notice there is still some commented out code and a TODO item. I am guessing those are there for future use.

PR #827, opened 2023-10-04
ZFIN-8830: Update NCBI Load Report

rtaylorzfin:
- The bottom of the report contains the most important information. Move it up to top.
- Remove the part with “------GenPept accessions with ZFIN genes still attributed to non-load publication ----------”
- Add gene symbols wherever gene IDs are shown


cmpich (approved):
Looks good.

PR #831, opened 2023-10-09
ZFIN-8432 make full 16 high-level term list show up in facet

cmpich:


source/org/zfin/search/service/FacetBuilderService.java
```diff
@@ -358,13 +367,18 @@ private List<FacetGroup> buildPhenotypeFacetGroup() {
         //facetGroups.add(buildSingleFacetGroup("Genes With Altered Expression","genes_with_altered_expression",false,  filterQuerySelectionMap));
 
         FacetGroup genotype = buildSingleFacetGroup("Genotype", FieldName.GENOTYPE_FULL_NAME.getName(),
-                Category.PHENOTYPE.getFacetQueriesForField(FieldName.GENOTYPE),
-                false);
+            Category.PHENOTYPE.getFacetQueriesForField(FieldName.GENOTYPE),
```
rtaylorzfin, line 370: instead of doing this in 3 places, we could create a convenience method

PR #833, opened 2023-10-10
ZFIN-8599: fix MO abbrev

rtaylorzfin:
Amy pointed out that the correct prefix for Morpholinos is MO and not MRPHLNO


cmpich (approved):
Looks good.

PR #835, opened 2023-10-11
ZFIN-8868: Fixes for convert_marker_type

rtaylorzfin:


cmpich (approved):
Remind me where expression_experiment table gets updated with the new marker ID?

PR #839, opened 2023-10-17
ZFIN-8868

cmpich:


rtaylorzfin (approved):

rtaylorzfin (changes requested):
looks like we need to remove the schema from temp tables. liquibase complained about it:

```
     Reason: liquibase.exception.DatabaseException: ERROR: cannot create temporary relation in non-temporary schema
```

rtaylorzfin (approved):

PR #842, opened 2023-10-18
ZFIN-8886 suppress records without start or end coordinate, update en…

cmpich:
…d coordinate for one allele a304


rtaylorzfin (approved):
Looks good

PR #845, opened 2023-10-19
Add newer postgresql client to compile image

rtaylorzfin:


zfinryanm (changes requested):
Security-wise, it would be preferable to download the key and add it to the source tree to be checked against rather than downloading it everytime.

rtaylorzfin:
@zfinryanm Good suggestion. I changed the Dockerfile and included the key in this commit

zfinryanm (approved):
Looks good.

PR #853, opened 2023-10-27
ZFIN-8599: Fix remaining issue with adding morpholino

rtaylorzfin:
I fixed a couple more places where we need to get the marker type based on "MO" not on "MRPHLNO"


rtaylorzfin:
@zfinryanm Hope you don't mind doing another code review. It would be nice to get this one on trunk since it addresses an issue Amy found while bailing 1149

cmpich (approved):
Looks good.

PR #854, opened 2023-10-30
ZFIN-8756: fix error message "files.forEach is not a function"

rtaylorzfin:
This fixes the console error that was complaining about "file.forEach". In this case "files" was not an Array, but a FileList so we have to convert it to an Array to use forEach.  Here are some screenshots to give context for what happens when trying to drag an image to the component:

Screenshot 2023-10-30 at 10 23 38 AM
Screenshot 2023-10-30 at 10 24 35 AM
Screenshot 2023-10-30 at 10 24 44 AM
Screenshot 2023-10-30 at 10 24 10 AM


cmpich (approved):
Looks good.

PR #856, opened 2023-11-03
ZFIN-8854: Show all construct name history

rtaylorzfin:
Some former names of constructs are in marker_history table, but not data_alias table. This combines those names on the construct view page


source/org/zfin/marker/presentation/PreviousNameLight.java
```diff
@@ -154,4 +102,23 @@ public static boolean isUninformative(String name){
         if(name.contains(":") && false==transgenicPrefixPattern.matcher(name).matches()) return true ;
         return false ;
     }
+
+    public void setPureAliasName(String pureAliasName) {
```
cmpich, line 106: I would not overwrite these two setters but rather overwrite the getter
getAlias ()
and include the business logic of adding italicized strings.

cmpich (approved):
Looks good.

PR #857, opened 2023-11-03
ZFIN-8707 load ctd publication

cmpich:


rtaylorzfin (approved):
Looks good. It would be nice if the line endings for source/org/zfin/datatransfer/ctd/PublicationCtd* could be fixed.

PR #858, opened 2023-11-03
ZFIN-8756: Add tooltips for figures that cannot be deleted.

rtaylorzfin:


home/javascript/react/containers/FigureEdit.js
```diff
@@ -22,6 +22,51 @@ function FigureEdit({ pubId }) {
             });
     }, [pubId]);
 
+    useLayoutEffect(() => {
```
cmpich, line 25: Can you not use useEffect()? useLayoutEffect() can slow down rendering...

rtaylorzfin: Updated

cmpich (approved):
Looks good.

PR #859, opened 2023-11-06
ZFIN-8854: Update to exclude aliases that differ from existing name by spaces only

rtaylorzfin:
As per request on ZFIN-8854 ticket


source/org/zfin/db/postGmakePostloaddb/1150/ZFIN-8854.sql
```diff
@@ -26,6 +26,8 @@ WHERE mhist_mrkr_zdb_id LIKE '%CONSTRCT%'
     NOT IN (SELECT dalias_data_zdb_id, dalias_alias FROM data_alias)
   AND (mhist_mrkr_zdb_id, mhist_mrkr_name_on_mhist_date)              -- filter out the current name from marker table
     NOT IN (SELECT mrkr_zdb_id, mrkr_name FROM marker WHERE mrkr_type LIKE '%CONSTRCT')
+  AND (mhist_mrkr_zdb_id, replace(mhist_mrkr_name_on_mhist_date, ' ', ''))  -- filter out alias that differs from current name only by spaces
```
cmpich, line 29: What if the space is in the middle of a multi-word alias? I assume you are trying to deal trailing spaces only...

cmpich: Ok, should be fine.

cmpich (approved):
Looks good.

PR #861, opened 2023-11-08
ZFIN-8418 refactor Expression Repository to be more compliant with Hi…

cmpich:
…bernate 5.x

### DIFF
--- a/source/org/zfin/expression/repository/HibernateExpressionRepository.java
+++ b/source/org/zfin/expression/repository/HibernateExpressionRepository.java
@@ -1566,10 +1566,9 @@ public class HibernateExpressionRepository implements ExpressionRepository {
 	public List<ExpressionResult> getExpressionsWithEntity(GenericTerm term) {
 		String hql = "select distinct expression from ExpressionResult expression where " +
 			"(entity.superterm = :term OR entity.subterm = :term) " +
-			" AND expressionFound = :expressionFound ";
+			" AND expressionFound = true ";
 		Query<ExpressionResult> query = HibernateUtil.currentSession().createQuery(hql, ExpressionResult.class);
 		query.setParameter("term", term);
-		query.setParameter("expressionFound", true);
 		return query.list();
 	}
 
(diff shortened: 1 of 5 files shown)


source/org/zfin/expression/repository/HibernateExpressionRepository.java
```diff
@@ -63,1372 +59,1296 @@
...
+
+	/**
+	 * Retrieve a single expression structure by ID.
+	 *
+	 * @param zdbID structure ID
+	 * @return expression structure
+	 */
+	public ExpressionStructure getExpressionStructure(String zdbID) {
+		Session session = HibernateUtil.currentSession();
+		return session.get(ExpressionStructure.class, zdbID);
```
rtaylorzfin, line 1319: Is there any potential for edge cases where ".get" returns null and ".uniqueResult" throws exception?

source/org/zfin/expression/repository/HibernateExpressionRepository.java
```diff
@@ -1462,437 +1382,434 @@ public void deleteExpressionResultPerFigure(ExpressionResult2 result, Figure fig
...
+	 * Retrieve Expressions for a given term.
+	 *
+	 * @param term term
+	 * @return list of expressions
+	 */
+	@Override
+	public List<ExpressionResult> getExpressionsWithEntity(GenericTerm term) {
+		String hql = "select distinct expression from ExpressionResult expression where " +
+			"(entity.superterm = :term OR entity.subterm = :term) " +
+			" AND expressionFound = :expressionFound ";
```
rtaylorzfin, line 1569: It's personal preference, but I like to embed the "true" in the hql myself, so it reads "... AND expressionFound = true ..."

There are a few other places in the refactor that contain query.setParameter("someTerm", true) too.

source/org/zfin/expression/repository/HibernateExpressionRepository.java
cmpich: Not fully sure but:
returning a null should be ok. the calling method should handle the null.
and the exception being thrown when uniqueResult will retrieve multiple records is fine as is. If the result set should always be unique this exception will fastest show if it is not true in some cases. I definitely would not call uniqueResult() knowing that there are cases where it is not true. In that case, you should call list() and then check according to best business logic how to handle the violation. if you need to check if something has a unique resultset you better call list() and then return the boolean and then proceed with both logical pathways from there. You typically should not use exceptions to go one or the other way in your code.

source/org/zfin/expression/repository/HibernateExpressionRepository.java
cmpich: I agree! If a parameter is fixed / hard-coded then do it right in the hql. Only use the setParameter() method if it gets passed in and can have multiple values.

PR #871, opened 2023-11-27
ZFIN-8809: convert ZDB-LINCRNAG-050208-65 to gene

rtaylorzfin:


cmpich (commented):
I cannot see the ZFIN-8809.sql file

cmpich (approved):
Looks good.

PR #875, opened 2023-11-28
ZFIN-8896 move more xenograft papers into right bin

cmpich:


source/org/zfin/db/postGmakePostloaddb/1150/ZFIN-8896.sql
rtaylorzfin, line 1: This looks good to me. A couple ways that could simplify it if you feel like making the effort:

This:
```
  and ((lower(title) ilike  '%xenograf%') OR (lower(title) ilike  '%xenotranspl%'))
```

Could omit the `lower` call since you're already using case-insensitive `ilike`:
```
  and (title ilike  '%xenograf%' OR title ilike  '%xenotranspl%')
```

Also, since you've already inserted into pubs_temp, you could re-use that table for the updates part, eg:
```
update pub_tracking_history
set pth_location_id = 19
where exists(
              select *
              from pubs_temp
              where zdb_id = pth_pub_zdb_id
                and pth_location_id in (6,10)
          )
;
```

cmpich: Great comments. Thanks.

PR #876, opened 2023-11-28
ZFIN-8801: Ensure RefSeq #s cannot be entered in as GenBank #s

rtaylorzfin:
…anks or other DB accessions


source/org/zfin/sequence.hbm.xml
```diff
@@ -90,6 +85,24 @@
             <one-to-many class="org.zfin.sequence.DisplayGroupMember"/>
         </set>
 
+        <set name="validationRules" table="reference_database_validation_rule" inverse="true" cascade="all">
+            <key column="fdbcvr_fdbcont_zdb_id"/>
+            <one-to-many class="org.zfin.sequence.ReferenceDatabaseValidationRule"/>
+        </set>
+    </class>
+
+    <class name="org.zfin.sequence.ReferenceDatabaseValidationRule" table="foreign_db_contains_validation_rule">
```
cmpich, line 94: It would be nice to have this new class use HIbernate annotations rather than xml definition.

rtaylorzfin: Okay, fixed

PR #889, opened 2023-12-15
ZFIN-8759: Replace angular with react for QuickFigure dialog

rtaylorzfin:
Replace angular for the feature shown here:

Screenshot 2023-12-15 at 12 22 37 PM


cmpich (approved):
Looks good.

PR #899, opened 2023-12-27
ZFIN-8949 convert ExpressionExperiment to ExpressionExperiment2 class

cmpich:


source/org/zfin/framework/presentation/tags/CreateLinkTag.java
```diff
@@ -182,7 +182,7 @@ else if (o instanceof Term) {
         else if (o instanceof TermDTO)
             link = TermDTOPresentation.getLink((TermDTO) o);
         else if (o instanceof ExpressionResult)
```
rtaylorzfin, line 184: Seems like this line should also be changed to ExpressionResult2.  Could even use the new syntax (https://blogs.oracle.com/javamagazine/post/pattern-matching-for-instanceof-in-java-14)

source/org/zfin/expression/repository/HibernateExpressionRepository.java
```diff
@@ -59,1296 +59,1283 @@
...
+            AND expressionFigureStage.startStage = :start
+            AND expressionFigureStage.endStage = :end
+            AND superTerm = :superTerm
+            AND expressionFound = :expFound
+            """, ExpressionResult2.class);
+        query.setParameter("expressionExperiment", result.getExpressionFigureStage().getExpressionExperiment());
+        query.setParameter("start", result.getExpressionFigureStage().getStartStage());
+        query.setParameter("end", result.getExpressionFigureStage().getEndStage());
+        query.setParameter("superTerm", unspecified);
+        query.setParameter("expFound", result.isExpressionFound());
```
rtaylorzfin, line 1173: I'm sure you've worked out the logic here, but I'm not sure why the `"expressionFound", true` gets changed to `"expFound", result.isExpressionFound()`

rtaylorzfin (approved):
Looks good

source/org/zfin/expression/repository/HibernateExpressionRepository.java
cmpich: Thanks for this catch. I will undo this change from the original one.

PR #901, opened 2023-12-30
ZFIN-8934: Add liquibase to merge zf2099 and zf3188

rtaylorzfin:
Using liquibase:

- update any foreign key references to the old ID so they point to the new ID.
- if an update would result in duplicate data violating a key constraint, just delete the row instead of updating it
- delete original feature 
- add zdb_replaced_data to redirect

### DIFF
--- a/lib/DB_functions/get_id_and_insert_active_data.sql
+++ b/lib/DB_functions/get_id_and_insert_active_data.sql
@@ -1,0 +1,10 @@
+-- Return the next available id for a given zobjtype and insert a new active data record
+CREATE OR REPLACE FUNCTION get_id_and_insert_active_data(obj_name varchar) RETURNS text AS $$
+DECLARE
+    next_id varchar;
+BEGIN
+    next_id := get_id(obj_name);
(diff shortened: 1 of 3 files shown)


lib/DB_functions/get_id_and_insert_active_data.sql
```diff
@@ -0,0 +1,10 @@
+-- Return the next available id for a given zobjtype and insert a new active data record
+CREATE OR REPLACE FUNCTION get_id_and_insert_active_data(obj_name varchar) RETURNS text AS $$
```
cmpich, line 2: Nice utility function!

cmpich (approved):
Looks good.

lib/DB_functions/get_id_and_insert_active_data.sql
rtaylorzfin: Haha, yeah sometimes it's the little things. There are probably some sql scripts in our codebase that could be made more readable with this. I was thinking if we need to get_id and reference it 3 or more times in a script, we can use session variables (https://blog.mclaughlinsoftware.com/2019/09/28/session-variables/)

PR #907, opened 2024-01-05
ZFIN-8937: Fix the next zf line number for "Add New Feature"

rtaylorzfin:
Account for feature tracking table when calculating the next zf line number for the "Add New Feature" interface in the curation UI.

### DIFF
--- a/source/org/zfin/feature/repository/HibernateFeatureRepository.java
+++ b/source/org/zfin/feature/repository/HibernateFeatureRepository.java
@@ -517,30 +517,52 @@ public class HibernateFeatureRepository implements FeatureRepository {
      * @return
      */
     public String getNextZFLineNum() {
-        return getNextZFLineNumWithoutFeatureTrackingCollision();
+        return getNextLineNumberForLabPrefix(FeaturePrefix.ZF);
+    }
+
+    public String getNextLineNumberForLabPrefix(String labPrefix) {
+        return getNextLineNumberForLabPrefix(getFeaturePrefixByPrefix(labPrefix));
+    }
+
+    public String getNextLineNumberForLabPrefix(FeaturePrefix labPrefix) {
+        return getNextLineNumberForLabPrefixWithoutFeatureTrackingCollision(labPrefix);
     }
 
     /**
-     * Get the next zf line number. First look at the feature table and add 1 to the max value.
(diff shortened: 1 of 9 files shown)


source/org/zfin/feature/repository/HibernateFeatureRepository.java
```diff
@@ -512,13 +512,37 @@ public List<FeaturePrefix> getLabPrefixes(String labName) {
...
+        }
+        return String.valueOf(nextLine);
     }
 
+    /**
+     * Get the next zf line number by looking at the feature table and adding 1 to the max value.
+     * @return next line number
+     */
+    private Integer getNextZFLineNumFromFeatureTable() {
+        String sql = "select max(cast(coalesce(feature_line_number,'0') as integer)) + 1 from feature where feature_lab_prefix_id = 194";
```
cmpich, line 542: I there a more generic fix? Lab-independent

rtaylorzfin: I updated the code to be lab independent. I noticed a couple of things:

1. The other labs don't use this logic to figure out the next line number. Maybe we should add that? Here's the frontend code:
```
        if (view.labDesignationBox.getSelected().equals(ZF_PREFIX)) {
            FeatureRPCService.App.getInstance().getNextZFLineNum(
...
```
2. Maybe related to #1, but the feature_line_number column of the feature table that we use for that calculation is often not an integer number like the ones for ZFIN.  Some examples: sagn025a, sagn028c, 001tk, 255H, iy121, a5, et33e20. It might be hard to come up with a generic way of finding the next "number" for those.

cmpich: Could you check with Yvonne or Holly if it should only apply to that one lab? I agree it's not possible when line numbers contain alphanumeric characters.

rtaylorzfin: I got the response in Jira:

> @Ryan Taylor no I do not think we want this for other labs as often multiple labs share a designation and have a certain range (Lab A uses numbers 1-100, Lab B uses numbers 1001-200, etc). We can get requests out of sequence and that makes it hard to manage. We do this for the zf numbers since we do not share our designation.

cmpich (approved):
Looks good. Make sure to only apply the automatic line number calculation for the zf lines.

PR #908, opened 2024-01-09
Fix failing test by skipping it for the next 8 months. 

rtaylorzfin:
We have a case to fix it (https://zfin.atlassian.net/browse/ZFIN-8955).


cmpich (approved):
Looks good.

PR #910, opened 2024-01-10
ZFIN-8797: Add smoke test to assert new STRs show up in blast results.

rtaylorzfin:
Add smoke test to detect recurring issue where recently added STRs don't show up in blast results.
The blast DB load logic is in another git repository (see https://zfin.atlassian.net/browse/ZFIN-8979) so that fix was performed in that repository. This PR adds a smoke test that can detect if the issue pops up again.


source/org/zfin/marker/SequenceFeature.java
```diff
@@ -13,33 +12,8 @@ public class SequenceFeature {
     @JsonView({View.API.class, View.ExpressedGeneAPI.class, View.UI.class})
     public String zdbID;
 
-    public String getNameOrder() {
```
cmpich, line 16: Did you add a lombok @Data or @Getter and @Setter?

cmpich (approved):
Looks good.

source/org/zfin/marker/SequenceFeature.java
rtaylorzfin: The "@Getter" and "@Setter"  were already there, but some of the methods were redundant.

PR #912, opened 2024-01-11
ZFIN-8940 ZFIN-8867 add three assembly chromosome, start and end attributes to download file, remove alliance submission files from registry

cmpich:
…download file


rtaylorzfin (approved):
Looks good

PR #914, opened 2024-01-11
ZFIN-8992: Create a skeleton for the react-based construct tab

rtaylorzfin:
Create a skeleton for the react-based construct tab in the curation interface and put it behind a feature flag.


cmpich (approved):
Looks good.

PR #917, opened 2024-01-18
ZFIN-8937: Bugfix - fetching a FeatureTracking object without related Feature

rtaylorzfin:
Fixes corner case where Hibernate tries to load the FeatureTracking's related Feature but the Feature doesn't exist.


cmpich (approved):
Looks good.

PR #920, opened 2024-01-18
ZFIN-8950: Use $PWD instead of $WORKSPACE 

rtaylorzfin:
ZFIN-8950: Use $PWD instead of $WORKSPACE because the behavior of $WORKSPACE is inconsistent between servers.

### DIFF
--- a/server_apps/jenkins/jobs/UniProt-Secondary-Term-Load/config.xml
+++ b/server_apps/jenkins/jobs/UniProt-Secondary-Term-Load/config.xml
@@ -34,7 +34,7 @@
   <blockBuildWhenUpstreamBuilding>false</blockBuildWhenUpstreamBuilding>
   <triggers>
     <hudson.triggers.TimerTrigger>
-      <spec></spec>
+      <spec/>
     </hudson.triggers.TimerTrigger>
   </triggers>
   <concurrentBuild>false</concurrentBuild>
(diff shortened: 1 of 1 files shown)


server_apps/jenkins/jobs/UniProt-Secondary-Term-Load/config.xml
```diff
@@ -34,7 +34,7 @@
   <blockBuildWhenUpstreamBuilding>false</blockBuildWhenUpstreamBuilding>
   <triggers>
     <hudson.triggers.TimerTrigger>
-      <spec></spec>
+      <spec/>
```
cmpich, line 37: This is not going to work if it should get scheduled with a cron expression. I assume this job is always meant to be run manually?

rtaylorzfin: Yes, I intend it to only be run manually. Perhaps we could configure it to run automatically at some point if we feel comfortable with that.

rtaylorzfin: In order to get this onto trunk, I'm going to merge this into main. I created another pull request against release-1152 so we can continue the discussion there if you have any more concerns before approving.

PR #922, opened 2024-01-20
ZFIN-8950: Bugfix: Report summary fails on dry run

rtaylorzfin:
Report summary fails when we do a dry run -- because we can't do a direct DB comparison of before and after. I added a check for this.


cmpich (approved):
Looks good.

PR #924, opened 2024-01-21
Uniprot Secondary Load Hotfix for Missing Go Term

rtaylorzfin:
Load was failing when the mapping file for ec2go was mapping to a go term that we don't have loaded in our DB. This patch skips those terms.


cmpich (approved):
Looks good. 
It does not add the missing eco-go mapping entries, though.

PR #925, opened 2024-01-22
ZFIN-8711: Remove all angular code since it is no longer used

rtaylorzfin:
Final removal of angular files and angular dependency in package.json


cmpich (approved):
Nice!

PR #928, opened 2024-01-24
ZFIN-9014: UniProt Secondary Term Load: Fix 2 issues

rtaylorzfin:
- marker_go_term_evidence inserts and deletes need to account for the "inferred_from" field (the interpro/ec/uniprot-kw accession)
- db_link removal needs to account for multiple uniprots per gene


cmpich (approved):
Looks good.

PR #932, opened 2024-01-25
ZFIN-8968: Add artifact-diff jenkins plugin

rtaylorzfin:
Include links to the diff pages in the emails that NCBI load sends out


cmpich (approved):
Looks good.

PR #934, opened 2024-01-26
ZFIN-9017: Keep orgID when creating person fails due to validation error

rtaylorzfin:
Fix case where admin creates a person from the lab page, but forgets to fill in required field, then gets redirected to page with validation errors and thus loses the organization information.


cmpich (approved):
Looks good.

PR #937, opened 2024-01-29
ZFIN-9029 handle NPE on antibody page when expression result is null

cmpich:


source/org/zfin/antibody/AntibodyService.java
```diff
@@ -629,81 +629,83 @@ public List<AnatomyLabel> getAntibodyDetailedLabelings() {
                 Set<ExpressionResult2> results = getExpressionResult2s(exp);
 
                 // loop through the set of ExpressionResult objects to get the related data
-                for (ExpressionResult2 result : results) {
-                    if (result.isExpressionFound()) {
+                if (results != null) {
```
rtaylorzfin, line 632: What do you think about changing getExpressionResult2s so that it never returns null and instead returns an empty set? That would allow us to not do null checks like this one and fix any other places that might have NPEs (like on line 540).  Example:

```
    private static Set<ExpressionResult2> getExpressionResult2s(ExpressionExperiment2 experiment) {
        if (CollectionUtils.isNotEmpty(experiment.getFigureStageSet())) {
            return experiment.getFigureStageSet().stream()
                .map(ExpressionFigureStage::getExpressionResultSet)
                .flatMap(Collection::stream)
                .collect(Collectors.toSet());
        }
        return Collections.emptySet();
    }
```

cmpich: Good suggestion. Had to check that other consumers of this method are dependent on null return value.

rtaylorzfin (approved):
Looks good

PR #938, opened 2024-01-31
ZFIN-8972 Create report of ZFIN genes with transcripts that are missing longest transcrtip from ensembl

cmpich:


package-lock.json
rtaylorzfin, line 1: Do we need the package-lock.json in this PR?

rtaylorzfin (approved):
Looks good

package-lock.json
cmpich: Not really, but it should go into the source tree, right? It's removing angular.
I am having trouble undoing this file in my commit history.

rtaylorzfin: No problem. We can commit it. Maybe we can talk about a better strategy for handling this file later.

cmpich: Yes, that file is frequently causing trouble.

PR #939, opened 2024-02-01
ZFIN-9035: Fix null pointer exception

rtaylorzfin:
Bugfix: when we converted from `.uniqueResult()` we didn't account for the case where result could be null.


zfinryanm (approved):
Looks fine to me.

PR #941, opened 2024-02-05
ZFIN-9030: GenesInvolvedIndexer load with stateful session

rtaylorzfin:
GenesInvolvedIndexer load needs to use stateful session in order to pull in related tables.
Also, some general cleanup.


cmpich (approved):
Looks good.

PR #959, opened 2024-02-20
ZFIN-9044 & ZFIN-9085: Fix ZDB ID lookup by uniprot accession & improve efficiency

rtaylorzfin:
This previously got ZFIN IDs from the richSequenceAdapter (ie. the uniprot release file), but it should be getting them from the context (ie. current ZFIN database contents). This commit fixes that.

I also added more efficient DB loading for the tables marker_to_protein, protein, protein_to_interpro.


cmpich (approved):
Looks good.

PR #960, opened 2024-02-22
ZFIN-8960 & ZFIN-8961: Put the same file in multiple categories

rtaylorzfin:
The current xml handler doesn't support this for the same filename, so we have to generate a copy of the file with a different name.


cmpich (approved):
We'll refactor the download's architecture in the near future and will deal with this issue then.

PR #964, opened 2024-02-27
ZFIN-8887: Replace Index-Faceted-Search_d with 3 child jobs 

rtaylorzfin:
Job Downstream Dependencies:
- Step 1 triggers Step 2 unless it fails (regen part of the indexer jobs)
- Step 2 triggers Step 3 unless it fails (solr index build)
- Step 3 runs cleanup code

Retries:
- Step 1 is retried on failure
- Step 2 is retried on failure
- Step 3 is not retried on failure (cleanup can happen on future runs)


zfinryanm (approved):
Looks good.

PR #971, opened 2024-03-06
Add experiment conditions for alliance phenotype file

cmpich:
We only use a high-level slim for conditionClassId and otherwise conditionId


rtaylorzfin (approved):
Looks good

PR #982, opened 2024-03-15
ZFIN-9109 limit the pagination record number to 10,000

cmpich:


rtaylorzfin (approved):

source/org/zfin/search/presentation/SearchPrototypeController.java
```diff
@@ -177,6 +177,11 @@ public String viewResults(@RequestParam(value = "q", required = false) String q,
 
         query.setRows(rows);
         int start = (page - 1) * rows;
+        //allow for not paginating beyond record number 10_000
+        if (start > 10000) {
```
rtaylorzfin, line 181: I just had a thought here. Do we want to log anything in these cases?

PR #985, opened 2024-03-18
ZFIN-9112: Remove "Format into a printable listing" button

rtaylorzfin:
Remove buttons for "Format into a printable listing" that generate a printable list of citations for a given publication search due to expensive back-end solr query that is causing problems in solr.

Redirect to faceted search for "List all pubs" link.

Redirect to faceted search for any direct requests to the endpoints for "Format into a printable listing"


cmpich (approved):
Looks good.

source/org/zfin/publication/presentation/PublicationSearchController.java
```diff
@@ -88,29 +89,15 @@ public String showSearchForm(Model model,
     @RequestMapping(value = "/search/printable", method = RequestMethod.GET)
```
rtaylorzfin, line 89: Change to apache redirect

PR #994, opened 2024-03-27
ZFIN-9122: Fix URL for gene2vega download file

rtaylorzfin:
It looks like the domain "ftp.ncbi.nih.gov" no longer works. Instead we should use "ftp.ncbi.nlm.nih.gov".


zfinryanm (approved):
Looks good.

PR #998, opened 2024-03-29
NCBI URL Update: all instances of ftp.ncbi.nih.gov to include .nlm

rtaylorzfin:


cmpich (approved):
Looks good. Some of the changes I already did so it may cause some merge conflict.

PR #1002, opened 2024-04-08
ZFIN-9123: In order to keep a history of gene2accession.gz and RefSeq…

rtaylorzfin:
…Catalog.gz, remove the wildcard delete (so as to keep the gene2accession.danio.<hash>.gz for example).


zfinryanm (approved):
Looks good but we'll want to keep an eye on disk space usage.

PR #1005, opened 2024-04-09
ZFIN-9152: UniProt logic bugfix for add/remove attribution

rtaylorzfin:
The logic for determining which gene/uniprots match up for load before deleting those that no longer match up, should include those actions of type "ADD_ATTRIBUTION" (ie. there is an existing manual attribution and we are adding an attribution to this load).


cmpich (approved):
Looks good.

PR #1007, opened 2024-04-10
ZFIN-9065 bugfix for DownloadFiles.sql

rtaylorzfin:
Move definition of temp table experiment_condition_with_zeco_and_chebi to before first use


cmpich (approved):
Looks good. Needs to go into release-1155 as well.

PR #1009, opened 2024-04-10
ZFIN-9119: Use the printf instead of print to get %.3f to get filled in

rtaylorzfin:
Fixes our gaf file headers that were previously created as:

```
!gaf-version: 2.2
!Version: %.3f
0.001!date-generated: 2024-04-04
!generated-by: ZFIN
!
```


cmpich (approved):
Looks good.

PR #1012, opened 2024-04-12
Fix build.gradle handling of perl module file and template strings

rtaylorzfin:
- Add task to deploy the ZFINPerlModules.pm file
- Use a filter to only apply the template logic to specific files (syntax like: `<!--|KEY|-->`)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -216,6 +216,143 @@ def translateVariableSet = [
         'WEBHOST_FASTA_FILE_PATH',
 ]
 
+//These are the only files that need to be searched and replaced (<!--|KEY|-->)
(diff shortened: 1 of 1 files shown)


rtaylorzfin:
I tweaked the logic for filling in placeholders in files that contain them.
The placeholders look like  in various source files.
The previous code was searching for placeholders in every file in server_apps/.
I changed it to only search for placeholders in files that we know need them.
I got that list of files by grepping through the source files for '<!--|'.

zfinryanm (approved):
Looks good to me.

cmpich (commented):
Looks good.

PR #1015, opened 2024-04-17
Bugfix: Use the correct temp view

rtaylorzfin:
This was using the base view before adding chemicals to it.


cmpich (approved):
Looks good.

PR #1016, opened 2024-04-17
ZFIN-9018

cmpich:
1. Refactor the LoadCTD to include mappings for cases where a chebi term has two cas references but only one of them is used in the mesh-cas file.
2. crreate export script that exports the mapping (unique) into a file according to SSSOM which means to add a predicate column and a column for the justification.


rtaylorzfin (approved):
Looks good!

PR #1017, opened 2024-04-17
ZFIN-9021: Patch: Moved the WHERE conditions around dates ...

rtaylorzfin:
to the inner query instead of the outer query where the dates have been truncated already.


cmpich (approved):
Looks good.

PR #1020, opened 2024-04-20
ZFIN-9169 reafactor linkML base class to use schema version, use modE…

cmpich:
…ntityId, agm_identifier


rtaylorzfin (approved):
Looks good

PR #1021, opened 2024-04-24
ZFIN-9142: Fix reports for genbank accessions in ncbi load

rtaylorzfin:
Also, fix data to revert to older version's attributions.


cmpich (approved):
Looks good.

PR #1025, opened 2024-04-26
ZFIN-9173: Remove axis and confluence soap jar

rtaylorzfin:
Remove everything that was using axis-1.4.jar (by way of confluence-soap-v2.jar). It was used to update the confluence pages through SOAP, but we don't do that anymore.


cmpich (approved):
That's a lot of cleanup. Thanks.

PR #1026, opened 2024-04-26
Replace commons-configuration-1.6.jar with a minimized version

rtaylorzfin:
Replace commons-configuration-1.6.jar with a minimized version that was generated using proguard. This removes unneeded classes from commons-configuration-1.6.jar and gets rid of the dependency on javax.servlet.

The better solution would be to fix the dependency on commons-configuration-1.6.jar to instead use [commons-configuration2](https://mvnrepository.com/artifact/org.apache.commons/commons-configuration2/2.10.1). But the dependency is through commons-configuration-ant-task-0.9.6.jar which appears to be an internal project of ZFIN, but I'm not sure where to find the source.

Proguard config (config.pro) was:
```
-injars commons-configuration-1.6.jar
-outjars commons-configuration-1.6-minimized.jar

# Keep only the specific classes required
-keep class org.apache.commons.configuration.ConfigurationException {
    public protected *;
}

-keep class org.apache.commons.configuration.PropertiesConfiguration {
    public protected *;
}

-keep class org.apache.commons.configuration.interpol.ConfigurationInterpolator {
    public protected *;
}

-keep class org.apache.commons.configuration.AbstractConfiguration {
    public protected *;
}

-keep class org.apache.commons.configuration.AbstractConfiguration$* {
    *;
}

-dontwarn

# Optional: enable shrinking and optimization
# -dontshrink
-dontoptimize
-dontobfuscate
```
And executed with `proguard-7.4.2/bin/proguard.sh @config.pro`


rtaylorzfin:
Closing this to try a different approach involving upgrading the intermediate package for commons-..ant...0.9.6.jar

PR #1027, opened 2024-04-29
ZFIN-9169 Fix link ml base

cmpich:
Use ModInternalId instead of modEntityID
start creating a dashboard for the Alliance submission files.


rtaylorzfin (approved):
Looks good

PR #1029, opened 2024-04-29
ZFIN-9173: Remove commons-logging jar

rtaylorzfin:
Remove our explicit use of the commons-logging-1.1.1.jar.


cmpich (approved):
Looks good.

PR #1030, opened 2024-04-30
adjust scheduling of faceted search job (Tuesday through Saturday and…

cmpich:
… removing scheduling of Alliance file submission. Needs always be manual


rtaylorzfin (approved):
Looks good. I assume the "Generate-Alliance-Files" one is removed since it is downstream of the downloads job.

PR #1032, opened 2024-04-30
Fix warnings about type coercion in sql files

rtaylorzfin:
I'm getting these noisy messages on gradle tasks:

psql:...create_stg_name_long.sql:50: NOTICE:  type reference stage.stg_name%TYPE converted to text
...

image


cmpich (approved):
Looks good.

PR #1037, opened 2024-05-06
Update Download-Files_d config.xml remove redundancy

rtaylorzfin:
The touch file command is redundant to the curl command. They both notify tomcat of the changes to the downloads directory.


cmpich (approved):
Looks good.

PR #1043, opened 2024-05-14
ZFIN-9058: Remove spaces on pasted sequences

rtaylorzfin:
Allow transforming text when pasted into the STR sequence input.


cmpich (commented):
Is there added validation on the server side to make sure it does not get into the database?

rtaylorzfin:
> Is there added validation on the server side to make sure it does not get into the database?

Yes, I tested sending a space to the backend with curl and it gets caught by the error handler here:

https://github.com/zfin/zfin/blob/main/source/org/zfin/marker/presentation/SequenceTargetingReagentAddController.java#L103

cmpich (commented):
Look good.

PR #1046, opened 2024-05-15
ZFIN-9032: Remove old uniprot load files since we have migrated to java

rtaylorzfin:


cmpich (approved):
Nice work. A lot of files are gone now!

PR #1052, opened 2024-05-17
ZFIN-9155 ZFIN-9156

cmpich:


source/org/zfin/search/Category.java
```diff
@@ -52,7 +52,7 @@ public enum Category {
             SOURCE,
             HOST_ORGANISM
     ),
-    ANATOMY("Anatomy / GO",
+    ANATOMY("Ontologies",
```
rtaylorzfin, line 55: Maybe change ANATOMY too

cmpich: Ok, I will change this when this in the main branch.

PR #1056, opened 2024-05-22
ZFIN-9207: Fix sorting on curator dashboard

rtaylorzfin:
The refactor in commit db5dd8516eb changed the sort order. It was previously sorted by "status" and that got changed to "status.name" (where status refers to PublicationTrackingStatus class). I believe sorting by "status" is equivalent to sorting by "status.id" and I confirmed that sort matches our requirements for this ticket.


cmpich (approved):
Looks good.

PR #1057, opened 2024-05-22
ZFIN-9194: Switched /home/informix urls to /research/zusers/informix …

zfinryanm:
…in ontology loads.


rtaylorzfin (approved):
Looks good

PR #1063, opened 2024-05-26
ZFIN-9170 remove gBrowse reference, remove INFORMIX references as well

cmpich:
Ryan T.: Could you check with some of the env variables in the 
source/org/zfin/properties/ZfinPropertiesEnum.java
Did you mean to include them in the general properties files?


rtaylorzfin:
@cmpich Are you looking at something specific?

rtaylorzfin (approved):

PR #1068, opened 2024-05-28
ZFIN-9076: Add ordering to include all columns on PDF check job

rtaylorzfin:
so the sort will be consistent across runs


cmpich (approved):
Looks good.

PR #1076, opened 2024-06-05
ZFIN-9237 & ZFIN-9238 & ZFIN-9118: Fix QuickSearchDialog

rtaylorzfin:
Fix 2 aspects of the facet dialog (QuickSearchDialog) that appears when you click "Show All" on the facets:

1. pagination could get into a weird state if you paginated, then changed the page size (ZFIN-9118)
2. Some facets have html (like <sup>...</sup>) that was not getting rendered correctly.


cmpich (approved):
Looks good.

PR #1077, opened 2024-06-07
ZFIN-9233: Don't limit knockdown download files to GENE types

rtaylorzfin:
Remove restriction on knockdown download files that limits the results to only GENE types.

Refactor download perl and sql files to remove
 templating syntax.


cmpich (approved):
Looks good.

PR #1078, opened 2024-06-07
ZFIN-9243: Give a better error message if there is an email collision.

rtaylorzfin:
Give a better error message if there is an email collision.


cmpich (approved):
Looks good.

PR #1079, opened 2024-06-10
Fix-Test: getAllEnsemblGenes

rtaylorzfin:
Fix test to match method signature change in getAllEnsemblGenes


cmpich (approved):
Looks good.

PR #1080, opened 2024-06-10
Clean temp tables for marker_go_term_evidence

rtaylorzfin:
We can delete these temp tables after running.


cmpich (approved):
Looks good.

PR #1083, opened 2024-06-12
ZFIN-9194: Fixing ECO load. Seems that ant changed and needs the full…

zfinryanm:
… path of the executable now.


cmpich (approved):
Looks good.

PR #1085, opened 2024-06-14
ZFIN-9247: Incremental improvement to construct edit

rtaylorzfin:
These changes allow editing constructs. I expect we will want to get some feedback and refine it further.


cmpich (approved):
Looks good.

PR #1087, opened 2024-06-17
HotFix: Cannot claim publications due to null value in PublicationTrackingController.java

rtaylorzfin:
Add null check for status


zfinryanm (approved):
Looks good.

PR #1095, opened 2024-07-02
ZFIN-9294: Error message for adding fluorescent protein

rtaylorzfin:
Show a more informative error message when a fluorescent protein isn't in our DB yet:

Screenshot 2024-07-02 at 2 37 43 PM

### DIFF
--- a/source/org/zfin/fish/presentation/EfgAPIController.java
+++ b/source/org/zfin/fish/presentation/EfgAPIController.java
@@ -9,7 +9,9 @@ import org.zfin.marker.Marker;
 import org.zfin.marker.fluorescence.FluorescentMarker;
 import org.zfin.marker.fluorescence.FluorescentProtein;
 
+import javax.servlet.http.HttpServletResponse;
 import java.util.List;
+import java.util.Optional;
 
 import static org.zfin.repository.RepositoryFactory.getMarkerRepository;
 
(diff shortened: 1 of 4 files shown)


source/org/zfin/fish/presentation/EfgAPIController.java
```diff
@@ -9,45 +9,64 @@
...
 
 @RestController
 @RequestMapping("/api/efg")
 public class EfgAPIController {
 
+    //Create nested structure for errors such that js can call: responseJSON.fieldErrors[0].message (for example)
     @JsonView(View.API.class)
-    public record Fpbase(String fpId, String abbreviation) {
-    }
+    private record FpbaseDTO(String fpId, String abbreviation, FpbaseResponseFieldError[] fieldErrors) {
```
cmpich, line 24: Why did you change the name to FpbaseDTO? To indicate it is not a business object? I wonder if we should have a naming convention for record's.

rtaylorzfin: I tend to think of these value objects as being pretty much the same as DTO objects.  Its purpose is to represent the fpbase entity for transmitting for the server to the client--so that seems to fit the definition of DTO to me. If you prefer a different name, that's fine with me.

cmpich: So you want to make it clear that this object only serves the purpose of data transfer and thus make it clear in the class name? I can agree to that.

cmpich (approved):
Looks good.

PR #1098, opened 2024-07-08
ZFIN-9276: Merge Markers

rtaylorzfin:
- Fix the recursive walk through foreign keys to handle different schemas
- Add DB HOST for use in case this script runs on a different host than the DB
- Account for the possibility that a data alias may already exist before creating a new one


cmpich (approved):
Looks good.

PR #1103, opened 2024-07-18
ZFIN-9237: Use unique key for each row returned by API

rtaylorzfin:
Use JSON.stringify to create a unique key for each row based on the exact string representation of the row


cmpich (approved):
Looks good.

PR #1104, opened 2024-07-22
ZFIN-9118: QuickSearchDialog - recalculate page size on filter change

rtaylorzfin:
Fix issue where the filter changing causes a need for the number of pages to be recalculated.


cmpich (approved):
Looks good.

PR #1110, opened 2024-08-02
ZFIN-9239: Table layout

rtaylorzfin:
In some cases we do not want a fixed table (table-layout: fixed;). In order to accommodate long construct names with no breaking characters, this makes the Mutations/Transgenics table non-fixed.


cmpich (approved):
Looks good.

PR #1118, opened 2024-08-09
ZFIN-9285: Construct Edit React Form

rtaylorzfin:
Added the functionality recommended by curators to allow editing cassettes instead of needing to remove and re-add.
Update backend to be able to process both construct edits and construct adds.


cmpich (approved):
Looks good.

PR #1130, opened 2024-08-20
ZFIN-9303: Fix syntax error

rtaylorzfin:


server_apps/data_transfer/zfishbook/zfishbook.pl
```diff
@@ -154,14 +154,15 @@ ($)
 
 print "\n\nStarting to load ...\n\n\n" if $doTheLoad > 0;
 
-if ($doTheLoad > 0)[
```
cmpich, line 157: I am not sure what the difference is.
Just indentation?

rtaylorzfin: For some reason, the old code was using a square bracket '[' instead of curly braces '{'.  It's a syntax error so I'm not sure how it could have ever worked.

cmpich (approved):
Looks good.

PR #1131, opened 2024-08-21
ZFIN-9335: Add ensembl load pub as a UniProt "load pub" 

rtaylorzfin:
...meaning its attributions can be removed during uniprot load.

- cleanup: add artifact zipping to save space
- cleanup: add some null checks
- when looking for latest unprocessed release, don't consider any that are older than an existing processed release


rtaylorzfin:
Closing and combining with 9336

PR #1132, opened 2024-08-21
ZFIN-9335 and ZFIN-9336: Add ensembl load pub & log updates

rtaylorzfin:
ZFIN-9335: Add ensembl load pub as a UniProt "load pub" meaning its attributions can be removed during uniprot load.
ZFIN-9336: Log changes in the updates table so they can be viewed on history page

cleanup: add artifact zipping to save space
cleanup: add some null checks
when looking for latest unprocessed release, don't consider any that are older than an existing processed release


cmpich (approved):
Looks good.

PR #1134, opened 2024-08-22
ZFIN-9338: Fixes to construct edit functionality

rtaylorzfin:
- Not all constructs are of type 'Transgenic Construct'
- Filter list of constructs by pub to only include those in construct relationships by pub


cmpich (approved):
Looks good.

PR #1136, opened 2024-08-26
Perl libs deploy

rtaylorzfin:
This is meant to fix the issue with the perl module file copy getting stuck in a loop. I tested it locally with success, but it's hard to know since I can't duplicate the issue from PROD locally.


build.gradle
```diff
@@ -291,8 +291,7 @@ def whitelistTemplateFiles = [
         '**/pushToZirc.sql',
         '**/pullFromZIRC.pl',
         '**/pushToZirc.pl',
-        '**/orphanChecks.pl',                                       //server_apps/DB_maintenance
-        '**/fx_permission_check.pl',
+        '**/fx_permission_check.pl',                                //server_apps/DB_maintenance
```
zfinryanm, line 294: Should fx_permission_check.pl be here rather than one line down under DB_maintenance?

zfinryanm (commented):
Can you look at the comment for build.gradle?

build.gradle
rtaylorzfin: Hmm, maybe my comments in this part of the code aren't very clear. I had intended the path on the right to indicate that the file on that line and all files below it (until the next comment) are in the same directory.

For example, in the following screenshot, the files `loadNewSNPs.sql` and `loadNewSNPAttrs.sql` are both in the same directory.

Screenshot 2024-08-29 at 4 18 23 PM

Since I removed orphanChecks.pl from the list of files that need the template processing, I had to move that comment down to the next file in the same dir (fx_permission_check.pl).

 I think I'm being consistent with the comments in this PR, but perhaps I should rearrange these comments to be more clear?

zfinryanm: Ah, I see. Word wrap made me misunderstand.

zfinryanm (approved):
Looks good.

PR #1140, opened 2024-08-30
ZFIN-9206 make ensembl first choice for transcript records, add havan…

cmpich:
…na-exists checkbox


home/WEB-INF/tags/markerTranscriptDataTable.tag
```diff
@@ -9,6 +9,7 @@
             <th>Type <a class="popup-link info-popup-link" href="/action/marker/transcript-types"></a></th>
             <th>Name</th>
             <th>Annotation Method</th>
+            <th width="10%">Has Havanna Data</th>
```
rtaylorzfin, line 12: I believe Havana has only one "n"

rtaylorzfin (approved):
Looks good

PR #1151, opened 2024-09-10
ZFIN-9282: Phenotype-Statistics-Report_m

rtaylorzfin:
- Use environment variables instead of `<!--|...|-->`
- Fix join logic:
Previously, the sql logic for "Number of Genes with TALEN phenotype in a Tg background (TALEN used transiently)" was failing to include a join condition for joining fish_str and fish tables resulting in a cross join. I refactored it to explicitly use join operations to make the logic easier to read.


cmpich (approved):
Looks good.

PR #1159, opened 2024-09-20
ZFIN-9370: Upgrade jackson annotations. 

rtaylorzfin:
See this for more info: https://stackoverflow.com/a/76638838


cmpich (approved):
Looks good. I assume the newer library is already in the repository.

rtaylorzfin:
> Looks good. I assume the newer library is already in the repository.

Yes, that's correct

PR #1161, opened 2024-09-21
ZFIN-9081: Change updates view

rtaylorzfin:
- Include publication history updates in the updates view
- Change to react-based page for updates view
- Revert changes to the updates table that are redundant to the publication_history updates

Screenshot:
image


cmpich (approved):
Looks good.

PR #1164, opened 2024-09-24
ZFIN-9217 create blast database from full Ensembl transcript one to o…

cmpich:
…nly contain records that exist in ZFIN


rtaylorzfin (approved):
Looks good

PR #1166, opened 2024-09-25
Convert HQL to Text Blocks

rtaylorzfin:


cmpich (approved):
Looks good.
Should we agree on a standard SQL query notation in terms of newlines?
Something for the CS meeting.

rtaylorzfin:
> Looks good. Should we agree on a standard SQL query notation in terms of newlines? Something for the CS meeting.

Sounds good. I added it to the agenda. Perhaps we could use the pg_format tool.

PR #1172, opened 2024-10-03
ZFIN-9295: (rework) Fix the backend server logic to trim whitespace

rtaylorzfin:
Also, run the load monthly.


zfinryanm (approved):
Looks good.

PR #1188, opened 2024-10-21
Image Service Unit Tests: If the CONVERT_BINARY_PATH doesn't exist, c…

rtaylorzfin:
…heck for the convert binary in the path.


cmpich (approved):
Looks good.

PR #1195, opened 2024-10-29
ZFIN-9405: SEO: Handle error "Duplicate without user-selected canonical"

rtaylorzfin:


home/WEB-INF/tags/layout/page.tag
```diff
@@ -32,6 +32,10 @@
 
         <title>${title}</title>
 
+    <c:if test="${not empty canonicalUrl}">
+        <link rel="canonical" href="https://${zfn:getProperty("DOMAIN_NAME")}${canonicalUrl}" />
```
rtaylorzfin, line 36: Is there a better way?

rtaylorzfin: I moved this logic into the JSP getCanonical function

cmpich (approved):
Looks good.

PR #1197, opened 2024-10-29
ZFIN-9406: Handle 500 exceptions in search console reports

rtaylorzfin:
Most of the errors in the report for 500 exceptions were due to visits to our old gbrowse page (eg. /action/gbrowse/?name=ENSDARG00000005626). This commit returns a page with a link to the newer jbrowse page.


cmpich (approved):
Looks good.

PR #1204, opened 2024-11-01
ZFIN-9358: use zdbID string instead of object to avoid issues with passing in Feature object

rtaylorzfin:


cmpich (approved):
Looks good.

PR #1205, opened 2024-11-04
ZFIN-9431: Change genes to LNCRNAG

rtaylorzfin:
Also, add some liquibase boilerplate for the next few releases


cmpich (approved):
Looks good :)

PR #1209, opened 2024-11-06
ZFIN-9433: Change the logic for the phenotype indexer MultiChebiCondition

rtaylorzfin:
MultiChebiCondition is set to true for instances where there are multiple experiments and at least one of the experiments is Chebi-related.


cmpich (approved):
Looks good.

PR #1214, opened 2024-11-08
Hide features on ZFIN lab page

rtaylorzfin:
Hide this on ZFIN lab page. We used to have this logic before the new "prototype" pages were released (See: ZFIN-9082).

image


zfinryanm (approved):
Looks good.

PR #1216, opened 2024-11-09
Fix delete rule test

rtaylorzfin:
Fix is to postpone running the tests until after the next gpad load


test/org/zfin/infrastructure/delete/DeleteRuleTest.java
```diff
@@ -167,12 +175,16 @@ public void regionValidation() {
 
     @Test
     public void sTRValidation() {
+        //ignore this test until 2/1/25
```
rtaylorzfin, line 178: Let's see if the new gpad load fixes these before merging

PR #1221, opened 2024-11-14
ZFIN-9407: Use `unaccent` postgresql function to allow searches for authors with accented letters.

rtaylorzfin:


cmpich (approved):
Looks good. I did not know about unaccent!

PR #1230, opened 2024-11-20
ZFIN-9426 add three-letter code to two ECO codes.

cmpich:


rtaylorzfin (approved):
Looks good. I didn't know about eco_go_mapping. Nice

PR #1233, opened 2024-11-21
Update XmlSitemapIndexSet.java

rtaylorzfin:
Sitemap entries should use <sitemap> tag instead of <url> tag


zfinryanm (approved):
Looks fine.

PR #1240, opened 2024-11-25
Update prompt in bash

rtaylorzfin:
Example of before and after:

```
zfin.org:.../source_roots/zfin.org$
schlapp:.../source_roots/zfin.org$
```


cmpich (approved):
Looks good. Will you inherit those changes for your instance as well?

PR #1252, opened 2024-12-05
ZFIN-9405: Canonical computation moved from the jsp to the controller

rtaylorzfin:


cmpich (approved):
Looks ok. Need to test it.

PR #1259, opened 2024-12-11
Triple Upgrade: remove jackson-core-2.13.0-rc1.jar that conflicts with jackson-core-2.15.2.jar

rtaylorzfin:


cmpich (approved):
Looks good.

PR #1261, opened 2024-12-12
Remove liquibase 9281 from database history

rtaylorzfin:
This might not be the right fix, but it is a quick and easy way to remove ZFIN-9281.sql from the liquibase history.


rtaylorzfin:
Closing. We manually ran the SQL.

PR #1269, opened 2024-12-18
ZFIN-9438: Update country codes

rtaylorzfin:
Add country code to as many registered users as possible. Use matching logic based on lab memberships and also contents of "address" column.


cmpich (approved):
Looks good.

PR #1272, opened 2024-12-20
ZFIN-9062: Fix shifting layout on term page

rtaylorzfin:
Use the css property 'visibility' instead of 'display' so the hiding/showing logic doesn't shift things around.


cmpich (approved):
Looks good.

PR #1273, opened 2024-12-20
ZFIN-9445: Update links out to PDB IDs

rtaylorzfin:
Change the links out to PDB IDs.  Use a controller endpoint for presenting multiple PDB IDs if more than one is associated with a given UniProt ID.


source/org/zfin/infrastructure/presentation/PdbLinkController.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.stream.Collectors;
+
+import static org.zfin.repository.RepositoryFactory.getMarkerRepository;
+
+@Controller
+@RequestMapping("/infrastructure")
+@Log4j2
+public class PdbLinkController {
+
+    public static final String PDB_URL = "https://rcsb.org/structure/";
```
cmpich, line 21: Could you get this URL from the foreign_db record of the database?

rtaylorzfin: Good idea. I updated it. Thanks

cmpich (approved):
Looks good.

PR #1275, opened 2024-12-30
ZFIN-9490 convert from expression_experiment to expression_experiment2

cmpich:


rtaylorzfin (approved):
Looks good! Lots of cleanup!

PR #1279, opened 2025-01-03
Remove quotes from SOLR_HOME

rtaylorzfin:


zfinryanm (changes requested):
You will need to rebuild the properties files and check them in for this change to work.

PR #1291, opened 2025-01-15
ZFIN-9435: UniProt Load Report Improvements

rtaylorzfin:
Improve the reports for uniprot load including smaller output files.
Cross check deleted uniprot accessions against NCBI to see if there is an obsolete RefSeq involved in the mapping.


cmpich (approved):
Great work! I'll need a walk-through at another time.

PR #1292, opened 2025-01-22
ZFIN-9408: Add company ID to company page.

rtaylorzfin:
Make IDs copyable for antibody, company, lab, person views. Don't show an empty orcid link if there's no orcid.


cmpich (approved):
Looks good.

PR #1298, opened 2025-01-24
ZFIN-9028: Generate error message for "Known Insertion Site"

rtaylorzfin:
if a chromosome location is specified without marking as "Known Insertion Site"


cmpich (approved):
Looks good.

PR #1299, opened 2025-01-24
Fix Test: createPerson

rtaylorzfin:
Fix the ordering of AccountInfo and Person creation for new entity mapping

### DIFF
--- a/source/org/zfin/profile/service/ProfileService.java
+++ b/source/org/zfin/profile/service/ProfileService.java
@@ -527,25 +527,30 @@ public class ProfileService {
 
         person.setEmailPrivacyPreference(getDefaultEmailPrivacyPreference());
 
-        AccountInfo accountInfo = new AccountInfo();
         String login = person.getPutativeLoginName();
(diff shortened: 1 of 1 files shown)


source/org/zfin/profile/service/ProfileService.java
```diff
@@ -527,25 +527,30 @@ public Person createPerson(Person person) {
...
-        AccountInfo accountInfo = new AccountInfo();
         String login = person.getPutativeLoginName();
         if (StringUtils.isEmpty(login)) {
             login = person.getEmail();
         }
+
+        person.setDeceased(false);
+        person.setUrl(processUrl(person.getUrl()));
+
+        HibernateUtil.currentSession().save(person);
```
cmpich, line 538: Do you have to save the person record twice in this method?

rtaylorzfin: Good point. I removed it.

cmpich (commented):
Looks good.

PR #1300, opened 2025-01-24
ZFIN-9439: Set person's country based on organization

rtaylorzfin:


source/org/zfin/profile/service/ProfileService.java
```diff
@@ -527,25 +540,31 @@ public Person createPerson(Person person) {
...
-        AccountInfo accountInfo = new AccountInfo();
         String login = person.getPutativeLoginName();
         if (StringUtils.isEmpty(login)) {
             login = person.getEmail();
         }
+
+        person.setDeceased(false);
+        person.setUrl(processUrl(person.getUrl()));
+
+        HibernateUtil.currentSession().save(person);
```
cmpich, line 551: I still see two lines with
.save(person)
is that on purpose?

rtaylorzfin: Sorry, the original pull request was branched off my local code before I removed that redundant code. It is now fixed here.

PR #1301, opened 2025-01-24
ZFIN-9226: Show smaller set of years for annual stats by default.

rtaylorzfin:
By default only show the most recent 10 years. Give select boxes for choosing a different range.

Screenshot 2025-01-24 at 11 36 56 AM


cmpich (approved):
Looks good.

PR #1304, opened 2025-01-27
Fix test failure: Remove redundant test

rtaylorzfin:
Removed test doesn't provide any value that isn't already covered in ProfileServiceTest.java


cmpich (approved):
Looks good.

PR #1307, opened 2025-01-28
ZFIN-9303: Gradle task for zfishmeta load

rtaylorzfin:


cmpich (commented):
Does that need to be a jenkins job that needs to run regularly? It looks to me like a one-time data cleanup.

cmpich (approved):
Looks good.

PR #1311, opened 2025-01-29
ZFIN-9493 fix LB script, fix unit test

cmpich:


source/org/zfin/infrastructure/EnumValidationService.java
```diff
@@ -170,7 +170,9 @@ public void validateCurationRole() throws EnumValidationException {
     public void validateCurationLocation() throws EnumValidationException {
         String hql = "select distinct ptl_location_display from pub_tracking_location";
         List locationList = HibernateUtil.currentSession().createNativeQuery(hql).list();
```
rtaylorzfin, line 172: Would it be good to add a type to this, like `createNativeQuery(hql, String.class)...` ?

rtaylorzfin (approved):
Looks good

source/org/zfin/infrastructure/EnumValidationService.java
cmpich: Good idea.

cmpich: Changed it also in the other methods.

PR #1312, opened 2025-01-29
Fix "off-by-one" error in ScrollableResults

rtaylorzfin:


cmpich (approved):
Let's get it in and test the table paginations in the UI...

rtaylorzfin:
@cmpich 
This is an example from before the fix showing the difference between the same URL on main branch versus release-1165 on PROD:

shot1
shot2

cmpich:
So, it was a problem, a silent one.

PR #1314, opened 2025-01-30
Fix table annotation for table in ui schema

rtaylorzfin:


cmpich (approved):
Looks good.
If that works we have to make that change also on the other ui tables.

PR #1338, opened 2025-02-21
Disable nested search

rtaylorzfin:
Use dotfile that flags for jenkins to disable nested search. This seems to be the recommended way to disable:

see: https://github.com/jenkinsci/nested-view-plugin/compare/nested-view-1.23...nested-view-1.24
and: http://plugins.jenkins.io/nested-view/releases/#version_1.24


cmpich (approved):
Looks good.

PR #1339, opened 2025-02-22
ZFIN-9557: NCBI Gene Load -- create zdb_active_data entry at same time as pre-load table

rtaylorzfin:
Previous version:
```
update ncbi_gene_load set zdb_id = get_id('DBLINK');
...
insert into zdb_active_data select zdb_id from ncbi_gene_load;
```

For some reason, the insert line is causing a duplicate ID error. I want to try this approach of combining those 2 lines into a single line:
```
update ncbi_gene_load set zdb_id = get_id_and_insert_active_data('DBLINK');
```


cmpich (approved):
Looks good.

PR #1341, opened 2025-02-25
Use eclipse transformer tool to change jar from using javax to using jakarta

rtaylorzfin:
Example command for transforming:

```
java -jar org.eclipse.transformer.cli-0.5.0.jar blast-serialization-1.0.jar blast-serialization-1.0-eclipse-transformed.jar
```


cmpich (approved):
Looks great! Thanks for doing this.

PR #1344, opened 2025-02-28
Add some helpful output to the logger in case of issues with the blast directory permissions

rtaylorzfin:


cmpich (approved):
Sounds like a helpful utitility

PR #1346, opened 2025-02-28
ZFIN-9579: Display error on figure label update (1166)

rtaylorzfin:
Display error if curator tries to set figure label to the same as an existing label.


cmpich (approved):
Looks good.

PR #1354, opened 2025-03-05
ZFIN-9598 fix HQL to apply is not null constraint

cmpich:


rtaylorzfin (approved):
Looks good! I wonder if there are other instances of this hiding out

PR #1365, opened 2025-03-10
ZFIN-9616 & ZFIN-9615 & ZFIN-9620: Fix jenkins jobs (1166)

rtaylorzfin:
Error:
```
2025-03-09 09:18:48,453 [main] INFO  datatransfer.AbstractScriptWrapper - Executing: update journal set jrnl_isoabbrev = 'BMC Anesthesiol',  jrnl_medabbrev = 'BMC Anesthesiol' where jrnl_zdb_id='ZDB-JRNL-250307-2' ;
Exception in thread "main" org.hibernate.PropertyValueException: not-null property references a null or transient value : org.zfin.publication.Journal.publisher
```


server_apps/jenkins/jobs/Load-Complete-Author-Names_d/config.xml
```diff
@@ -17,6 +17,12 @@
     <concurrentBuild>false</concurrentBuild>
     <customWorkspace>$TARGETROOT</customWorkspace>
     <builders>
+        <hudson.tasks.Shell>
+            <command><![CDATA[
```
cmpich, line 21: My preference would be to create that 'target'-type of directory from Java so that the execution is less dependent on the environment in which it is executed. Applies to the following similar changes as well.

rtaylorzfin: @cmpich , Good point. I removed this type of logic from the xml and added it to java or shell scripts where necessary.

cmpich (approved):
Looks good.

PR #1381, opened 2025-03-20
ZFIN-8989: Update social media links

rtaylorzfin:


home/css/layout.scss
```diff
@@ -278,6 +278,12 @@ footer {
         margin-right: 8px;
     }
 
+    .bluesky-icon {
```
cmpich, line 281: It would be nice to use font awesome to display the icon instead of svg.

rtaylorzfin: Updated to use the fontawesome icon

cmpich (approved):
Looks good.

PR #1390, opened 2025-04-03
ZFIN-9643: Fix duplicate data on Expression Summary page

rtaylorzfin:
Enforce uniqueness for ExpressedGene.expressionStatements by using a Set internally


source/org/zfin/expression/service/ExpressionService.java
```diff
@@ -658,7 +658,7 @@ public static List<FigureExpressionSummary> createExpressionFigureSummaryFromExp
                 }
                 Collections.sort(expressionStatements);
                 ExpressedGene expressedGene = new ExpressedGene(expressionExperiment.getGene());
-                expressedGene.setExpressionStatements(expressionStatements);
```
rtaylorzfin, line 661: This was the main culprit

cmpich (approved):
Looks good.

PR #1392, opened 2025-04-07
ZFIN-9627 create report for ensembl transcripts

cmpich:


server_apps/data_transfer/Downloads/GFF3/knockdown_reagents/load_knockdown_reagents.sh
```diff
@@ -5,7 +5,7 @@ TARGETDIR="$TARGETROOT/server_apps/data_transfer/Downloads/GFF3/knockdown_reagen
```
rtaylorzfin, line 5: I noticed this file uses `/bin/sh`. I wonder if we should be explicit about the shell and change it to /bin/bash? If we do, we can use `set -e` which will stop the script if it encounters any errors.

cmpich: Sounds good.

test/org/zfin/sequence/SequenceServiceTest.java
```diff
@@ -95,19 +96,61 @@ public void testSequenceServiceJsonAggregation() {
...
+                allGeneIds.add(values[0]);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        try {
+            FileWriter writer = new FileWriter("sequenceIDs.txt", false);
+
+
```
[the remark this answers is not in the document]
cmpich: Since this is only a unit test method no need to handle resource management.

rtaylorzfin (approved):
Looks good

PR #1397, opened 2025-04-09
ZFIN-9622: Handle edge case for manually curated UniProt accessions flagged for deletion

rtaylorzfin:
Handle edge case for manually curated UniProt accessions flagged for deletion

- Added logic to detect manually curated UniProt accessions that would otherwise be lost
	- Introduced new SubType: MANUALLY_CURATED_ACCESSION_WOULD_BE_LOST
	- Enhanced CheckLostUniprotsForObsoletesHandler to review this subtype
	- These accessions are included in the report for curator review
- Improved log messages
- Improve efficiency with reuse of reference database object


cmpich (approved):
Looks good.

PR #1403, opened 2025-04-16
ZFIN-9678: Use "ON CONFLICT" to avoid adding duplicates that would violate unique key

rtaylorzfin:


cmpich (approved):
Nice command in SQL.

PR #1408, opened 2025-04-23
ZFIN-9683 point zfa load to github zf-edit file

cmpich:


cmpich:
In the new Hibernate version you don't have to set the 'dialect' any longer.

rtaylorzfin (approved):

PR #1413, opened 2025-04-25
ZFIN-9674: Update fetchPubsFromPubMed to allow updating existing pub authors

rtaylorzfin:
Allow passing a fetchMode = 'update' to the fetchPubsFromPubMed.groovy script along with PubMed IDs to update existing rows in publication table with the values fetched from NCBI. It updates the authors column only right now.

Example:
```
groovy -cp "$GROOVY_CLASSPATH:." -DfetchMode=update ./fetchPubsFromPubMed.groovy 39821319
```

Also, fix the content-type header to be "UTF-8":
Example curl commands to inspect the content-type from tomcat:
```
docker exec -it -u root zfin_org-tomcat-1 curl -D - -s -o /dev/null -k https://localhost:8443/action/publication/ZDB-PUB-250118-4 | grep -i content-type:

docker exec -it -u root zfin_org-tomcat-1 curl -D - -s -o /dev/null -k https://localhost:8443/action/marker/gene/view/ZDB-GENE-990415-8 | grep -i content-type:
```

Example output:
```
Before change:
Content-Type: text/html;charset=ISO-8859-1

After change:
Content-Type: text/html;charset=UTF-8
```

### DIFF
--- a/server_apps/data_transfer/PUBMED/updateExistingPubs.sql
+++ b/server_apps/data_transfer/PUBMED/updateExistingPubs.sql
@@ -1,0 +1,105 @@
+begin work;
+
+create temp table tmp_pubs (
+  pmcid text,
+  mid text,
+  pmid integer,
+  keywords text,
+  title text,
+  pages text,
+  abstract text,
+  authors text,
+  numAuthors int,
+  year text,
+  month text,
+  day text,
+  issn text,
+  volume text,
+  issue text,
+  journaltitle text,
+  iso text,
+  status text,
+  pubType text);
+
+\copy tmp_pubs from 'parsePubs.log';
+
+delete from tmp_pubs
+where (authors = 'none' or authors is null)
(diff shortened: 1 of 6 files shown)


server_apps/data_transfer/PUBMED/updateExistingPubs.sql
```diff
@@ -0,0 +1,105 @@
...
+  volume text,
+  issue text,
+  journaltitle text,
+  iso text,
+  status text,
+  journal_zdb_id text,
+  pubType text
+);
+
+insert into tmp_new_pubs
```
cmpich, line 53: Do you need to drop this table at the end of this script?

rtaylorzfin: As a temp table, these should get dropped automatically.

cmpich (approved):
Looks good.

PR #1423, opened 2025-05-08
ZFIN-9669: Implement JMX-based metrics monitoring with CAPTCHA control

rtaylorzfin:
- Added MetricsMonitor to collect Tomcat JMX metrics and system load averages
- Created HighLoadAction to automatically enable/disable CAPTCHA under high load
- Updated Tomcat Dockerfile to enable JMX port for remote monitoring
- System triggers CAPTCHA when load or request processing time exceeds thresholds
- Added CSV-formatted metrics output for monitoring and analysis

Can run in a tomcat container like so:
```
docker exec -it zfin_org-tomcat-1 bash -lc 'cd /opt/zfin/www_homes/zfin.org/home/WEB-INF/classes && java -cp ".:../lib/*" org.zfin.task.metricsmonitor.MetricsMonitor'
```

### DIFF
--- a/source/org/zfin/task/metricsmonitor/HighLoadAction.java
+++ b/source/org/zfin/task/metricsmonitor/HighLoadAction.java
@@ -6,7 +6,7 @@ import java.util.Date;
 public class HighLoadAction {
 
     public static final String TRIGGER_ON_SQL = "update zdb_feature_flag set zfeatflag_enabled=true, zfeatflag_last_modified=now() where zfeatflag_name in ('Enable Captcha', 'Use altcha')";
-    public static final String TRIGGER_OFF_SQL = "update zdb_feature_flag set zfeatflag_enabled=false";
+    public static final String TRIGGER_OFF_SQL = "update zdb_feature_flag set zfeatflag_enabled=false, zfeatflag_last_modified=now() where zfeatflag_name in ('Enable Captcha')";
     private final MetricsMonitorConfig configuration;
 
     public boolean isTriggered = false;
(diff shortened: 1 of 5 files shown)


source/org/zfin/task/metricsmonitor/MetricsMonitor.java
```diff
@@ -0,0 +1,192 @@
+package org.zfin.task.metricsmonitor;
```
rtaylorzfin, line 1: We should probably run this as a thread in tomcat.

source/org/zfin/task/metricsmonitor/HighLoadAction.java
```diff
@@ -0,0 +1,67 @@
+package org.zfin.task.metricsmonitor;
+
+import java.sql.*;
+import java.util.Date;
+
+public class HighLoadAction {
```
rtaylorzfin, line 6: Clean up SQL when converting to a thread

source/org/zfin/task/metricsmonitor/MetricsMonitor.java
```diff
@@ -0,0 +1,192 @@
...
+    private SystemMetricsDTO collectMetrics() {
+        SystemMetricsDTO metricsRow = new SystemMetricsDTO();
+        metricsRow.setCurrentTime(timeFormat.format(new Date()));
+        collectResponseTimeMetrics(metricsRow);
+        collectLoadAverageMetrics(metricsRow);
+        return metricsRow;
+    }
+
+    private void printMetrics(SystemMetricsDTO metrics) {
+        if (rowNumber == 0) {
```
rtaylorzfin, line 90: Only print this out if log level is debug?
We could configure a specific file for it to write to.

rtaylorzfin: We should have a way to turn off this thread just in case something unexpected starts happening

cmpich (approved):
Looks good. The changes can go into a separate ticket.

PR #1440, opened 2025-05-22
ZFIN-9718 (part 2): Fix saving of evidence codes for orthologs

rtaylorzfin:


source/org/zfin/orthology/repository/HibernateOrthologyRepository.java
```diff
@@ -208,4 +208,25 @@ public List<NcbiOrthoExternalReference> getNcbiExternalReferenceList(String ncbi
         return query.list();
     }
 
+    @Override
+    public void saveEvidenceCode(OrthologEvidence evidence) {
+        String sql = """
+                    INSERT INTO ortholog_evidence (oev_ortho_zdb_id, oev_evidence_code, oev_pub_zdb_id, oev_evidence_term_zdb_id)
```
cmpich, line 214: Why do you use a native query for insertion rather than the Business object OrthologEvidence

rtaylorzfin: Good question! I would prefer to use the business object. When I changed the mappings from xml to annotations for OrthologEvidence, the database writes for OrthologEvidence stopped working. It appears to be due to the fact that the OrthologEvidence hibernate entity uses a composite key `OrthologEvidenceId`. 

OrthologEvidenceId is composed of:

        private String ortholog;
        private String evidenceCode;
        private String publication;

This works well for reading from the DB, but when writing, there is a mismatch between the types because evidenceCode is a string in the composite key, but it is an EvidenceCode object in the hibernate mapping.

I tried changing the type in the OrthologEvidenceId class, but that caused other issues. If you have a suggestion for a better solution, I would like to try it.

cmpich (approved):
Looks good for now. We should look into how to use composite keys with object type in hibernate mapping annotations.

PR #1453, opened 2025-06-09
ZFIN-9700: Run Load-Flank-Seq_w daily

rtaylorzfin:
- 6 of 7 days, only run for features that have been modified.
- Add fgmd_modified_at column to track when changes have been made


cmpich (approved):
Looks good.

PR #1501, opened 2025-07-18
icon-fix: do not display the external link indicator on blog lists

rtaylorzfin:
image

### DIFF
--- a/home/javascript/react/components/WikiBlogPostList.js
+++ b/home/javascript/react/components/WikiBlogPostList.js
@@ -56,7 +56,7 @@ class WikiBlogPostList extends Component {
                 <LoadingSpinner loading={loading} />
                 <div className='wiki-list-controls'>
                     <span>{ next && <a href='#' onClick={this.loadNext}>Load More</a> }</span>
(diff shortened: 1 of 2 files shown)


home/javascript/react/components/WikiBlogPostList.js
```diff
@@ -56,7 +56,7 @@ class WikiBlogPostList extends Component {
                 <LoadingSpinner loading={loading} />
                 <div className='wiki-list-controls'>
                     <span>{ next && <a href='#' onClick={this.loadNext}>Load More</a> }</span>
-                    <span>{ showAll && <a href={`https://${process.env.WIKI_HOST}${showAll}`}>See All <i className='fas fa-chevron-right' /></a>}</span>
+                    <span>{ showAll && <a className='no-icon' href={`https://${process.env.WIKI_HOST}${showAll}`}>See All <i className='fas fa-chevron-right' /></a>}</span>
```
cmpich, line 59: Do we have more than one wiki_host variable? I think we probably can just hard-code the wiki host to the one we have.

rtaylorzfin: Done

zfinryanm (approved):
Looks good. I doubt we will ever use a dev wiki in the future so the wiki host change is fine.

PR #1517, opened 2025-08-11
ZFIN-9874: Fix link in Mapping Details list

rtaylorzfin:
From the mapping details list, pass along the "source" information (eg. GRCz12tu, GRCz11, etc.)


zfinryanm (approved):
Looks okay to me but you might what Christian to look at it as well.

PR #1521, opened 2025-08-14
ZFIN-9851: Change default link to jbrowse to point to the version with GRCz12tu

rtaylorzfin:


zfinryanm (approved):
Looks good. Did you want to add a reminder to change it back like in the other PR?

PR #1531, opened 2025-08-19
ZFIN-9874: Hide z11 when we have z12 on mapping details page.

rtaylorzfin:


rtaylorzfin:
We discovered that deleting them causes some unintended consequences. Ultimately, we should delete them. I made this PR to see if we want to put in a short term fix that filters them out from the UI.

PR #1576, opened 2025-10-01
ZFIN-9958: Add tool for getting list of "not alive" NCBI GeneIDs and their replacements (if they exist).

rtaylorzfin:


cmpich (approved):
Looks good.

PR #1588, opened 2025-10-07
Trunk Test Failing: Fix failing test getAnatomyForMarker

rtaylorzfin:
I did some debugging and compared the SQL in the test to the SQL that gets generated by the method under test (getWildTypeAnatomyExpressionForMarker).

Here's a screenshot of the diff (after using pg_format):

image

You can see that the effective difference is that the hql query includes the qualifier "AND genox_is_standard = TRUE".  
I updated the test sql to include that clause and it fixes the test. That said, I'm not sure how valuable this test is since it seems to just test if the hql matches the expected sql--but it's not clear to me which one is the correct one if they differ.


zfinryanm (approved):
Looks fine.

PR #1665, opened 2026-01-09
ZFIN-10059: Instead of running "Load-GPAD-Noctua_w" every day, run "L…

rtaylorzfin:
…oad-GPAD-Noctua-Daily-Trigger_d" daily and it will trigger the former if necessary.


cmpich (approved):
Looks good, though it adds a new Jenkins Job as discussed.

PR #1681, opened 2026-01-22
ZFIN-10069 & ZFIN-10015 & ZFIN-10070: Remove Vega logic and add reports

rtaylorzfin:
ZFIN-10069: Remove references to vega load pub
ZFIN-10015: Add unlinked (no ncbi gene ID) gene count tracking to NCBIDirectPort
ZFIN-10070: Add report that shows the genes that have ncbi links that are marked as obsolete


cmpich (approved):
Looks good.

PR #1687, opened 2026-01-27
ZFIN-10030: Entity mapping for PLC classes in profile.hbm.xml

rtaylorzfin:
Some errors encountered on the way:

- org.hibernate.AnnotationException: Property 'org.zfin.profile.OrganizationFeaturePrefix.currentDesignation' belongs to an '@IdClass' but has no matching property in entity class 'org.zfin.profile.OrganizationFeaturePrefix' (every property of the '@IdClass' must have a corresponding persistent property in the '@Entity' class)
- Duplicate generator name zdbIdGenerator: Renamed the generators in Hit.java and ReferenceDatabase.java to zdbIdGeneratorForHit and zdbIdGeneratorForReferenceDatabase respectively, to ensure global uniqueness and avoid conflicts with other entities.
- MappingException for FishSupplier: Added FishSupplier, GenotypeSupplier, and FeatureSupplier to hibernate.cfg.xml. These were missing subclasses of ObjectSupplier.
- org.hibernate.exception.SQLGrammarException: JDBC exception executing SQL [select p1_0.zdb_id,... where p1_0.zdb_id=?] [ERROR: column c1_1.active does not exist Position: 60] [n/a]
- java.io.StreamCorruptedException: invalid stream header: 5A44422D
- revert whitespace changes


cmpich (approved):
Looks good.

PR #1692, opened 2026-01-28
Make Pub page mesh terms sort order deterministic 

rtaylorzfin:
by updating meshHeadings to use SortedSet for natural sorting.

Example of the random sort order between refreshes:

https://github.com/user-attachments/assets/3a040b4d-2f9d-4e40-96e8-17bcd346870a


cmpich (approved):
Looks good.

PR #1706, opened 2026-02-05
ZFIN-10113: Simplify properties files - single source of truth

rtaylorzfin:
This pull request refactors the build system to simplify property and filter management, removing legacy mechanisms and files in favor of direct usage of the main `zfin.properties` file. It also fixes a path typo and updates logic for environment-specific operations. The most important changes are grouped below:

**Build property and filter management simplification:**

* The build now loads properties directly from `${basedir}/home/WEB-INF/zfin.properties` instead of using instance-specific property files and removes checks for their existence. This change affects how properties are loaded and referenced throughout the build process.
* Filters for the build are now loaded directly from `zfin.properties` using a `<filterset>` in `build.xml`, eliminating the need to generate and import `buildfiles/allfilter.xml`. The legacy `exportAntTokenFilters` target is deprecated.
* Numerous macrodefs and targets for exporting properties to various formats (env, bash, TT, flattened, allfilter.xml) have been removed from `build.xml`, streamlining the build process.
* Remove commons-configuration-ant-task-0.9.7.jar

**Other fixes and improvements:**

* Fixed a typo in the `classes.dir` property in `buildfiles/base-properties.xml`, correcting the path from `WEB-INFclasses` to `WEB-INF/classes`.
* Updated the hostname check in `buildfiles/utilities.xml` to use the `INSTANCE` property instead.
*


cmpich (approved):
Looks good.

PR #1714, opened 2026-02-10
ZFIN-10082: Add gene symbol match CSV to NCBI match task and load report

rtaylorzfin:
Add a second CSV output (ncbi_gene_symbol_matches.csv) to NcbiMatchThroughEnsemblTask that identifies NCBI-ZFIN matches where the NCBI gene symbol matches marker_abbrev, filtered to only include rows without an existing NCBI db_link. Includes a zdb_ids_match column comparing NCBI's predicted ZDB ID to the marker's ZDB ID.

Also adds a report action in NCBIDirectPort that reads the new CSV and includes it in the NCBI load report.


cmpich (commented):
Could this be cleaned up to be more readable?

cmpich (approved):
Looks great!

PR #1715, opened 2026-02-11
ZFIN-10098: Merge journals

rtaylorzfin:
Update config.xml to use ZFIN_PROPERTIES_PATH instead of INSTANCE.env and enhance journal redirection logic.


cmpich (approved):
I assume the Jenkins job is doing the merge and that code did not need any modification?

PR #1725, opened 2026-02-13
ZFIN-10129: Add GRCz12tu liftover report 

rtaylorzfin:
Add a new Jenkins validation job to identify markers and features with coordinates on older assemblies (GRCv10, GRCz10, GRCz11, Zv9) that lack GRCz12tu coordinates. Also extend ReportGenerator to produce CSV and XLSX output alongside the existing HTML and TXT formats, using Apache Commons CSV and Apache POI (both already in our dependencies).


cmpich (approved):
Looks good.

PR #1727, opened 2026-02-16
ZFIN-10093: Add Jenkins job to report withdrawn NCBI Gene IDs and gen…

rtaylorzfin:
…es missing external IDs

Two reports in one job (Check-NCBI-Withdrawn-Gene-IDs):
1. Lists all discontinued zebrafish NCBI Gene IDs from gene_history, showing which ones ZFIN links to (old and replacement IDs).
2. Lists ZFIN genes missing NCBI Gene ID or UniProt associations, or flagged as not in current annotation release.


cmpich (approved):
Looks good.

PR #1728, opened 2026-02-17
ZFIN-10105: Make captcha cookie tamper-proof with HMAC signing

rtaylorzfin:
Replace the static cookie value (rcv_true) with an HMAC-SHA256 signed token bound to the session ID and timestamp. This prevents bots from trivially bypassing captcha by setting the cookie value directly.


cmpich (approved):
Looks good.

PR #1737, opened 2026-02-23
ZFIN-10137: Add retry with backoff and TokenStorage API key to NCBI calls

rtaylorzfin:
Replace hardcoded NCBI API key with TokenStorage lookup (NCBI_API_TOKEN) and add exponential backoff retry logic for transient HTTP errors (429, 5xx) across all NCBI API calls in PubmedUtils. PMC OAI/OA endpoints correctly omit the api_key parameter since they don't support it. Individual publication failures no longer abort the entire job.


cmpich (approved):
Looks good.

PR #1749, opened 2026-03-03
ZFIN-10106: Fix smoke tests to authenticate before accessing captcha-protected pages

rtaylorzfin:
Anatomy and Antibody smoke tests now extend AbstractSecureSmokeTest to log in before each test. AbstractSecureSmokeTest creates a unique test user with BCrypt-encoded random password, proper Hibernate back-references, and cleans up the user after each test.


cmpich (approved):
Looks great.

PR #1750, opened 2026-03-03
ZFIN-10106: Replace captcha boilerplate with @RequiresCaptcha interceptor

rtaylorzfin:
Add CaptchaInterceptor that detects @RequiresCaptcha on handler methods and performs the captcha redirect check before the controller runs. This eliminates identical 4-line boilerplate from 13 methods across 11 controllers. The two POST handlers in PersonSubmission/OrganizationSubmission retain their non-standard captcha behavior unchanged.


cmpich (approved):
Nice simplication!

PR #1757, opened 2026-03-05
ZFIN-10079: Feature detail page updates

rtaylorzfin:
**Feature validation and backend logic:**

* Added a backend validation to ensure that for an 'is allele of' relationship, the feature and gene must be on the same chromosome. If not, a validation exception is thrown.
* Introduced a new RPC method `getReferenceSequence` in `FeatureRPCServiceImpl` to fetch the reference sequence for a given assembly, chromosome, and location range. 
* Enhanced feature validation to require Chromosome, Assembly, Start/End Location, and Evidence Code when a Sequence of Variant is specified, and to prevent the Sequence of Variant from being identical to the Sequence of Reference.

**UI/UX improvements for feature curation:**

* The UI now automatically fetches and displays the reference sequence when all required fields are populated for relevant feature types (deletion, point mutation, indel, MNV), providing immediate feedback to the curator.
* The chromosome, assembly, and location fields dynamically update their visibility and labels based on the selected feature type, and location hints are shown contextually for deletions, indels, and insertions.
* For point mutations, the end location is automatically set to match the start location, both in the DTO and in the UI. 

**UI code and layout updates:**

* Refactored UI XML files to use `chromosomeLabel`, `chromosomePanel`, `startLocationLabel`, `endLocationPanel`, and contextual HTML hints.
* Updated `GenomicMutationDetailView` to use a `TextArea` for the sequence of reference, replacing the previous `StringTextBox`, to better support longer sequences.

**Minor code and logic improvements:**

* Removed an unnecessary dirty-check on the reference sequence in `EditMutationDetailPresenter`.
* Improved import statements and minor code cleanup.

These changes collectively enhance the accuracy, usability, and robustness of the feature curation workflow.

### DIFF
--- a/source/org/zfin/gwt/curation/server/FeatureRPCServiceImpl.java
+++ b/source/org/zfin/gwt/curation/server/FeatureRPCServiceImpl.java
@@ -23,6 +23,7 @@ import org.zfin.gwt.root.ui.ValidationException;
 import org.zfin.gwt.root.util.NullpointerException;
 import org.zfin.infrastructure.*;
 import org.zfin.infrastructure.repository.InfrastructureRepository;
+import org.zfin.mapping.FeatureGenomeLocation;
 import org.zfin.mapping.FeatureLocation;
 import org.zfin.mapping.GenomicLocationService;
 import org.zfin.marker.Marker;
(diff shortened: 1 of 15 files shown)


cmpich (approved):
Looks good.

PR #1758, opened 2026-03-05
ZFIN-10116: Universal CSV download for search results using injected …

rtaylorzfin:
…attributes

Replace per-category CSV transformers with a universal approach that reuses ResultService.injectAttributes() to include all columns shown on the search results page. HTML is stripped from attribute values and CSV escaping is handled by Apache Commons CSV.


cmpich (approved):
Looks good.

PR #1763, opened 2026-03-09
Clean up: Remove file that is generated by job

rtaylorzfin:
The jenkins job "Dump-RNACentral-File_w" creates this file. I don't think we want to keep it in our source tree.


rtaylorzfin:
@cmpich @zfinryanm These are all files that look like they can be cleaned up and removed from our repo. Do you know if we need any of these?

cmpich (approved):
Looks good.

PR #1765, opened 2026-03-09
ZFIN-10174: Convert marker.hbm.xml to annotation-based Hibernate mappings

rtaylorzfin:
Replace the marker.hbm.xml file with JPA/Hibernate annotations on 18 entity classes and 2 embeddable component classes.

Key changes:
- Marker uses @Inheritance(JOINED) with @PrimaryKeyJoinColumn on subclasses (Antibody, Clone, Transcript, SNP, SequenceTargetingReagent)
- SequenceFeature becomes @MappedSuperclass holding the @Id and ZdbIdGenerator configuration
- MarkerRelationship gets direct @Entity mapping, replacing the abstract discriminator hierarchy (which always resolved to one type)
- MarkerType uses @Access(PROPERTY) to preserve setter side effects that populate transient enum fields
- STRMarkerSequence and SNPMarkerSequence become @Embeddable
- MarkerType and MarkerTypeGroup retain @Immutable and @Cache(READ_ONLY)
- Custom enum types use @Type(StringEnumValueUserType.class)
- HQL reference to AbstractMarkerRelationshipInterface updated to MarkerRelationship in HibernateAntibodyRepository
- ProbeLibrary.tissue uses targetEntity=GenericTerm.class since the field type is the Term interface


cmpich (approved):
Looks good.

PR #1766, opened 2026-03-10
ZFIN-8174: Remove dead GO curation tab code

rtaylorzfin:
The GO curation UI was replaced by Noctua years ago. Remove the GWT editor classes, RPC caching wrapper, JSP, and test that were exclusively used by the now-removed GO curation tab.


cmpich (approved):
Looks good. Nice to see so many files go away!

PR #1771, opened 2026-03-12
ZFIN-10182: Minimize regen_term() lock duration to prevent deadlocks

rtaylorzfin:
regen_term() was holding an AccessExclusiveLock on all_term_contains for the entire duration of populating and indexing, blocking all reads (~9200 queued queries, up to 25 min wait). Now indexes, constraints, and self-records are built on the _new table before the swap, reducing the exclusive lock to just two fast DDL renames.

Old table is renamed with a timestamp suffix and truncated rather than dropped, consistent with the pattern used by regen_genox. Cleanup is handled by regen_cleanup_renamed_tables() via regenPhenotypeMartCleanup.

make_alltermcontains_indexes.sql is now idempotent (IF NOT EXISTS) since regen_term() builds indexes internally.

### DIFF
--- a/lib/DB_functions/Regen_genox/regen_cleanup_renamed_tables.sql
+++ b/lib/DB_functions/Regen_genox/regen_cleanup_renamed_tables.sql
@@ -6,9 +6,9 @@ DECLARE
     result text;
 BEGIN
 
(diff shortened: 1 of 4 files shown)


cmpich (approved):
Looks good.

PR #1775, opened 2026-03-13
ZFIN-10191: Replace mailhog with mailpit to capture outgoing emails on trunk and test

rtaylorzfin:
This replaces mailhog with mailpit. It also changes the SMTP_HOST for trunk and test to point to mailpit so that trunk and test should no longer send emails out to actual email inboxes. mailpit must be running to be able to read those emails (`docker compose up -d mailpit`).


cmpich (approved):
Nice!

PR #1776, opened 2026-03-13
ZFIN-10178: Remove outdated INFGRP, DBLINK, and LINK entities from ZDB-PUB-040304-6

rtaylorzfin:
Use liquibase script to remove entities.
Also, added a simple display for DBLINK entities that is only available to root users:

image


cmpich (approved):
Looks good.

PR #1783, opened 2026-03-24
ZFIN-10189 update all_terms_contains table update

cmpich:


server_apps/data_transfer/LoadOntology/handleRelationships.sql
```diff
@@ -202,3 +227,7 @@ delete from term_relationship
      	    	    	where term_id = term_ont_id
 			and term_zdb_id = termrel_term_2_zdb_id);
 
+!echo "Updating transitive closure for removed relationships";
+
+select update_all_term_contains_batch_remove();
```
rtaylorzfin, line 232: @cmpich 
Where is this defined? I don't see `update_all_term_contains_batch_remove.sql`

rtaylorzfin (changes requested):
@cmpich I think it's missing `update_all_term_contains_batch_remove.sql`

server_apps/data_transfer/LoadOntology/handleRelationships.sql
cmpich: thanks for pointing this out. I had to add it to git. Done.

rtaylorzfin (approved):
Looks good

PR #1787, opened 2026-03-25
ZFIN-10185 & ZFIN-10190 & ZFIN-10187: Remove permanent tables named '*tmp*' and replace with temp tables

rtaylorzfin:
ZFIN-10190: Remove ZDB-INFGRP records
ZFIN-10185: Clean up tables named tmpXYZ that should not stick around in the DB
ZFIN-10185: Remove informix-specific sql files
ZFIN-10187: Use proper "temp" tables in sql queries to guarantee they go away after session


cmpich (approved):
Looks nice!

PR #1795, opened 2026-04-01
ZFIN-10156: Construct edit feature add

rtaylorzfin:
The functional change is isolated to a single file in the first commit. 
The subsequent commits fix errors from the typescript compiler to make the code more strict.


home/javascript/react/components/data-table/index.d.ts
rtaylorzfin: This is a TypeScript declaration file (.d.ts) providing type definitions for the data-table module. The data-table component is likely written in plain JavaScript (JSX), so
  this file tells TypeScript what its exports look like without converting the source to .ts.

  It declares types for three components:
  - DataTable (default export) — table with columns, data URL, pagination, sorting, filtering, download options
  - DataList — similar to DataTable but without tableFixed (probably a different layout)
  - CollapseTable — a collapsible variant with a catch-all [key: string]: unknown for extra props

  And shared types:
  - ColumnDefinition<T> — defines a column's label, content renderer, width, filter name, visibility
  - TableState — pagination/sort/filter state
  - DownloadOption / SortOption — self-explanatory

  This file exists so that the TypeScript components (like the construct editor) can import from data-table without type errors, even though data-table itself remains
  JavaScript.

cmpich (approved):
Looks good.

PR #1803, opened 2026-04-03
ZFIN-10010: Fix duplicate genome location rows on mapping detail page

rtaylorzfin:
The mapping detail page showed duplicate rows in the genome browser table due to two issues:

1. markerAssemblyUpdate.sql inserted NCBILoader and ZFIN rows without checking for existing records, creating true duplicates across runs. Added NOT EXISTS guards to both inserts.

2. The UCSC loader creates one row per RefSeq accession per gene, which appear identical on-screen since accession isn't displayed. Added deduplication by (entity, chromosome, start, end, source, assembly) in MappingService.sortAndFilterGenomeBrowserLocations().

Includes a Liquibase cleanup script to remove existing NCBILoader and ZFIN duplicate rows.


source/org/zfin/db/postGmakePostloaddb/1181/ZFIN-10010.sql
```diff
@@ -0,0 +1,93 @@
+--liquibase formatted sql
+--changeset rtaylor:ZFIN-10010
+
+-- Clean up duplicate rows in sequence_feature_chromosome_location_generated
+
+-- First get a list of duplicates to delete:
+SELECT
```
rtaylorzfin, line 7: Let's attach this as a report to the ticket

rtaylorzfin: Done

cmpich (approved):
Looks good.

PR #1804, opened 2026-04-03
ZFIN-10205: Sync mrkr_name and mrkr_abbrev with construct_name

rtaylorzfin:
8 constructs had mrkr_abbrev (and/or mrkr_name) out of sync with construct_name, the canonical source of truth. This changeset updates both marker fields to match construct_name for any mismatched rows.


cmpich (approved):
Looks good.

PR #1806, opened 2026-04-03
ZFIN-9566: Upgrade react

rtaylorzfin:


package.json
```diff
@@ -59,10 +59,10 @@
     "postcss": "^8.4.33",
     "prop-types": "15.8.1",
     "qs": "^6.11.0",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
```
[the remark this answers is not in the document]
rtaylorzfin: Updated the upgrade notes

home/javascript/react/containers/MarkerEditCloneData.js
```diff
@@ -14,7 +14,8 @@ const CloneData = ({
     vectorList,
     digestList,
     polymeraseList,
-    clone,
```
rtaylorzfin, line 17: Why is this necessary? What is the error message?

package.json
```diff
@@ -5,6 +5,8 @@
   "repository": "https://github.com/zfin/zfin",
   "license": "BSD 3-Clause License",
   "devDependencies": {
+    "@testing-library/jest-dom": "^5.17.0",
```
rtaylorzfin: We could remove these

home/javascript/react/containers/MarkerEditCloneData.js
rtaylorzfin: @cmpich I took another look at the html on trunk to see what was happening before this change. The rendered html is definitely wrong:
```

<div class="__react-root"
--
id="MarkerEditCloneData"
data-clone-id='ZDB-BAC-030616-1'
data-clone=CLONE
zdbID: ZDB-BAC-030616-1
name: CH211-101N13
symbol: CH211-101N13
type: BAC
rating: null
problem: null
 
data-cloning-site-list='["Xb..."]'
data-polymerase-list='["Supe..."]'
data-clone-problem-type-list='["Chimeric",...]'
>
</div>
```

It turns out that the "data-clone" property was never used. Instead, it fetched the data it needed using fetch. I changed it to just remove that property.

cmpich (approved):
Looks great, thanks.

PR #1807, opened 2026-04-07
ZFIN-9723: Fix feature sort order not updating when abbreviation changes

rtaylorzfin:
Fix feature sort order update triggers for abbreviation and name changes.

The feature_abbrev and feature_name AFTER UPDATE triggers tried to set NEW.feature_abbrev_order and NEW.feature_name_order, but PostgreSQL ignores row modifications in AFTER triggers. This left stale sort-order values whenever a feature abbreviation or name was changed post-creation.

Fix: make the existing BEFORE INSERT trigger also fire on UPDATE so that scrub_char and zero_pad are always applied. Clean up the AFTER triggers to contain only side-effects (history logging, genotype name cascades).

Fix any existing stale sort-order values.
Trigger definitions are updated in lib/DB_triggers/ and applied via gradle make.

### DIFF
--- a/lib/DB_triggers/feature.sql
+++ b/lib/DB_triggers/feature.sql
@@ -42,7 +42,7 @@ END;
 $BODY$ LANGUAGE plpgsql;
 
 CREATE TRIGGER feature_before_trigger
-BEFORE INSERT ON feature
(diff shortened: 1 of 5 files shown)


cmpich (approved):
Looks good.

PR #1808, opened 2026-04-07
ZFIN-9793: allow list of IPs for captcha bypass

rtaylorzfin:
Add a REST endpoint that returns the original client IP as resolved through proxy headers (X-Forwarded-For, X-Real-IP) along with all request metadata and headers. Useful for verifying that the nginx proxy is correctly forwarding client IP information to tomcat.


cmpich (approved):
Looks great.

PR #1812, opened 2026-04-09
(release-1180 fix) ZFIN-10150: Fix getPDFandImages.groovy S3 migration issues

rtaylorzfin:
Use @Field annotation for DOWNLOADABLE_EXTENSIONS so it's accessible from closures inside def methods. Skip figure processing when image files aren't available on disk (non-open-access articles in S3 only have text files, not images).


cmpich (approved):
Looks good.

PR #1817, opened 2026-04-13
export db port on localhost by default or by environment variable

rtaylorzfin:


cmpich (approved):
Thanks for fixing this.

PR #1818, opened 2026-04-13
Proposal for Liquibase: include all

rtaylorzfin:
I keep running into merge conflicts when multiple PRs all touch the same changelog file (eg. `source/org/zfin/db/postGmakePostloaddb/1180/db.changelog.master.xml`). This change would allow us to put all our migrations in a subdirectory named `migrations` and liquibase will run all of them in alphabetical order. For example, our 1182 files could all go into `source/org/zfin/db/postGmakePostloaddb/1182/migrations`.


cmpich (approved):
Cool. I like this simplification.

PR #1819, opened 2026-04-14
ZFIN-10225: Fix closed ResultSet error in profile search methods

rtaylorzfin:
Replace ScrollableResults with getResultList() in searchLabs, searchCompanies, and searchPeople to avoid JDBC cursor conflicts caused by eager-loaded Person associations triggering secondary queries that close the open cursor on PostgreSQL.

The problem was a circular EAGER fetch chain:
1. searchLabs uses query.scroll() → returns a ScrollableResults backed by an open JDBC cursor
2. As Hibernate materializes each Lab, it eagerly loads Organization.owner (a Person, since @ManyToOne defaults to EAGER)
3. Person.labs is annotated @ManyToMany(fetch = FetchType.EAGER) at Person.java:142 — so loading the Person triggers loading all their Labs
4. This secondary query executes on the same JDBC connection while the ScrollableResults cursor is still open
5. PostgreSQL's JDBC driver can only have one open ResultSet per connection — the secondary query closes the original cursor
6. When scrollableResults.next() is called again → "closed ResultSet"

The fix is to replace query.scroll() with query.getResultList() in these search methods. The scroll-based approach provides no benefit here anyway — the code iterates through ALL rows to get a total count, so there's no memory savings from the cursor.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1201,6 +1201,7 @@ test {
             includeTestsMatching "org.zfin.httpunittest.MarkerViewSmokeTest"
             includeTestsMatching "org.zfin.sebservice.MarkerRestSmokeTest"
             includeTestsMatching "org.zfin.framework.presentation.HealthEndpointSmokeTest"
+            includeTestsMatching "org.zfin.profile.smoketest.ProfileSearchSmokeTest"
             //include specific method in any of the tests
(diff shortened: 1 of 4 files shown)


cmpich (approved):
Looks good.

PR #1823, opened 2026-04-21
ZFIN-10231 & ZFIN-10236: GAF load - accept multiple IEA inferences

rtaylorzfin:
GAF load: accept multiple IEA inferences, add error summary report

- Change IEA inference cardinality from exactly-1 to one-or-more. GOA now provides multiple inference references per annotation (e.g. UniRule:UR000376739|UniRule:UR000414636). These were being rejected, losing ~322 valid annotations per load.

- Log IEA annotations that load with multiple inferences so the change can be monitored.

- Add GafErrorSummary.java that parses the details file error section and generates a companion _error_summary.txt with categorized counts, gene-not-found deep dive (by ID type and annotation source), and examples for each error category. Runs automatically after details file is written.

Conveniences:

- Add Database-Load job (so we can run `gradle loaddb` through jenkins
- Add Error logging for issues relating to ncbi api key


cmpich (approved):
Looks great.

PR #1827, opened 2026-04-24
ZFIN-10193 & ZFIN-9280: fix pheno_term_fast_search regen null-PK and sequence ownership

rtaylorzfin:
CREATE TABLE ... AS SELECT ... WHERE FALSE copies column types but not defaults, so ptfs_pk_id in pheno_term_fast_search_tmp lost its nextval() default. Inserts left it NULL and ADD PRIMARY KEY failed. Restore the sequence default on the tmp column (regression of ba6a7e8631, dropped when #1799 rewrote the file).

Also reassign sequence ownership to the new live column after the rename-and-recreate swap in both pheno_term_regen.sql and refreshPhenotypeMart.sql. Otherwise the sequence remains OWNED BY the renamed _old_ table's column and is dropped when cleanup does DROP TABLE ... CASCADE, breaking the next regen.


cmpich (approved):
Looks good.

PR #1832, opened 2026-04-28
ZFIN-10074: paginate SPTrEMBL UniProt downloads

rtaylorzfin:
Replaced the wget-based /uniprotkb/stream call with curl-driven cursor pagination over /uniprotkb/search at size=500. The /stream endpoint was reproducibly truncating the zebrafish download with an SSL "unexpected eof" partway through, causing the Jenkins job to hang for hours; UniProt's docs explicitly recommend paginated /search for any large result set because each page failure is recoverable.

Also switches the query from organism_name:<name> (full-text) to organism_id:<taxid> (exact taxonomy match), and adds timestamped progress logging every 20 pages so a future stall is visible in the build log instead of silent.

The output filenames, sed >tr->sp rewrite, and downstream convertSPTrEMBL.sh / pushSPTrEMBL.sh contracts are unchanged.


cmpich (approved):
Looks good.

PR #1833, opened 2026-04-28
ZFIN-10269: Replace Ant-based ontofox/ROBOT step with gradle prepareZecoTaxonomyObo

rtaylorzfin:
Load-Zeco-Taxonomy-Ontology was the sole consumer of the create-obo-file Ant macro (which forked lib/Java/ontologies/robot-cleaned.jar). The fat robot-cleaned.jar had been stripped of its embedded jackson to avoid a NoSuchMethodError that the WAR no longer hits today, leaving ROBOT's CLI broken with NoClassDefFoundError on com/fasterxml/jackson/core/ObjectCodec.

Pull ROBOT 1.9.0 -> 1.9.10 from Maven Central via a dedicated `robotCli` gradle configuration that's isolated from runtime/implementation, so its classes never enter WEB-INF/lib. Two new tasks in console.gradle:
  - downloadZecoTaxonomyOwl: curl uploads the source.txt to ontofox and saves the returned .owl
  - prepareZecoTaxonomyObo: dependsOn the download; runs ROBOT convert via JavaExec against configurations.robotCli

The Jenkins job now invokes `gradle prepareZecoTaxonomyObo` directly (matching the pattern used by Check-And-Update-Journals_w and others) and the existing Ant load-zeco-taxonomy target is unchanged.

Removed:
  - server_apps/data_transfer/LoadOntology/build.xml: create-obo-file macro, create-zeco-taxonomy-owl-file target, create-go-qualifier-file target (no Jenkins job referenced it), and dead zecoTaxonomySourceURL / goQualifierSourceURL / goQualifierOboURL property defs
  - lib/Java/ontologies/robot-cleaned.jar (38 MB binary off the source tree)
  - jackson-datatype-guava implementation dep (only used by the bundled robot.jar; confirmed no source code references)

Verified end-to-end with `gradle prepareZecoTaxonomyObo`: produced zeco-taxonomy.obo (62.9 KB, 105 [Term] entries) at the path load-zeco-taxonomy reads.


cmpich (approved):
Looks great.

PR #1836, opened 2026-04-28
ZFIN-9922: Upgrade node to version 24 and upgrade packages

rtaylorzfin:


webpack.config.js
```diff
@@ -1,5 +1,4 @@
 const path = require('path');
-const { CleanWebpackPlugin } = require('clean-webpack-plugin');
```
rtaylorzfin, line 2: double check this one?

rtaylorzfin: The new "clean: true" option replaces the old CleanWebpackPlugin

zfinryanm (approved):
Looks good to me. Not sure if you want to get Christian's approval, as well.

PR #1837, opened 2026-04-29
ZFIN-10186 use Authorization only when a flag is set. By default do n…

cmpich:
…ot authorize, gets authenticated via IP address


rtaylorzfin (approved):
Makes sense!

PR #1838, opened 2026-04-29
ZFIN-10243: pull StudentMetricsJob roster from AccountInfo.student flag

rtaylorzfin:
Replaces hardcoded ZDB-PERS lookups with ProfileRepository.getStudents(), so the job picks up whichever submitters have is_student set without code changes.


cmpich (approved):
Looks good.

PR #1844, opened 2026-05-05
ZFIN-10245: Restore GRCz11 as a current assembly on the feature curation form

rtaylorzfin:
Re-add GRCz11 to the Add Feature dropdown, relax the server-side "must change to GRCz12tu" rule to also accept GRCz11, and run the flanking-sequence upsert against whichever of the two current assemblies is on the saved location.

### DIFF
--- a/source/org/zfin/gwt/curation/server/FeatureRPCServiceImpl.java
+++ b/source/org/zfin/gwt/curation/server/FeatureRPCServiceImpl.java
@@ -55,7 +55,6 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static org.zfin.framework.HibernateUtil.currentSession;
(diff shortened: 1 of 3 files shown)


cmpich (commented):
Looks good so far. I don't see the validation of the sequence against the true one from the genomic fasta file.

rtaylorzfin:
> Looks good so far. I don't see the validation of the sequence against the true one from the genomic fasta file.

@cmpich I believe you are referring to the code at FeatureRPCServiceImpl.java:1249-1267. specifically 1261 compares fgmd.getFgmdSeqRef() (the incoming
  sequence of reference) against the genome assembly via glService.getReferenceSequence(...) and throws "Sequence of Reference.

This code is unchanged. I deployed this code change on cell.zfin.org if you want to take a look

cmpich (approved):
Looks good.

PR #1845, opened 2026-05-05
ZFIN-10266 & ZFIN-10267 & ZFIN-10268: Zirc submission form

cmpich:


rtaylorzfin (approved):
Looks good. If you want to consult some of the feedback from copilot, or do it in another case. either is fine.

PR #1848, opened 2026-05-07
Zfin 10241

cmpich:


lib/DB_functions/update_all_term_contains_batch.sql
```diff
@@ -35,37 +36,51 @@ begin
...
+                                    alltermcon_min_contain_distance)
+    select container, contained, min(dist)
+    from (
+      select a.alltermcon_container_zdb_id  as container,
+             d.alltermcon_contained_zdb_id  as contained,
+             a.alltermcon_min_contain_distance + 1 + d.alltermcon_min_contain_distance as dist
+      from tmp_zfin_rels nr
+      join all_term_contains a on a.alltermcon_contained_zdb_id = nr.termrel_term_1_zdb_id
+      join all_term_contains d on d.alltermcon_container_zdb_id = nr.termrel_term_2_zdb_id
+      where nr.termrel_type in ('is_a', 'part_of', 'part of',
```
rtaylorzfin, line 68: we only want to use "is a" and "part of" and not "regulates" or variations of "regulates"

server_apps/DB_maintenance/regen.sh
```diff
@@ -11,11 +11,9 @@ echo 'select regen_genox(); ' | ${PGBINDIR}/psql -v ON_ERROR_STOP=1 ${DBNAME}
 echo "Starting regen_anatomy_counts at `date`"
 echo 'select regen_anatomy_counts();' | ${PGBINDIR}/psql -v ON_ERROR_STOP=1 ${DBNAME}
 
-echo "Starting regen_term at `date`"
-echo 'select regen_term();' | ${PGBINDIR}/psql -v ON_ERROR_STOP=1 ${DBNAME}
-
-echo "Starting regen_term_indexes at `date`"
-${PGBINDIR}/psql -v ON_ERROR_STOP=1 ${DBNAME} < ${SOURCEROOT}/server_apps/DB_maintenance/postgres/make_alltermcontains_indexes.sql
```
rtaylorzfin, line 18: Check if this can be dropped

lib/DB_functions/update_all_term_contains_batch.sql
cmpich: I checked the old regen_term script and that included those relationships as well. So, I'd leave it as is for now.

server_apps/DB_maintenance/regen.sh
cmpich: Two reasons, both apply:                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                     
  1. The script existed only as a paired step for regen_term(). Old behaviour: regen_term() dropped/rebuilt all_term_contains without indexes; cron then ran make_alltermcontains_indexes.sql to re-add the PK, the two indexes, and the two FKs. Once regen_term() is 
  removed from cron, that paired step has nothing to attach to.
  2. Even if regen_term() were still called, the script is now redundant inside it. The current function body (you can see it via pg_get_functiondef) builds the PK/indexes/FKs on the _new table before the swap and renames them to the canonical names after, so by 
  the time the swap completes, all_term_contains is fully indexed. Re-running make_alltermcontains_indexes.sql on top would only hit IF NOT EXISTS/if not exists in pg_constraint no-op branches.                                                                      
   
  So removing the line wasn't just a side-effect of dropping regen_term() — it was retired in step with the function's earlier refactor that pulled index creation inside the same transaction-equivalent flow (so the live table was never unindexed).                
                  
  The script file itself is kept for the standalone case (e.g. someone hand-builds the table from scratch), which is why its top-of-file comment says "no-op after regen_term(), or standalone if needed."

rtaylorzfin (approved):
Looks good

PR #1854, opened 2026-05-09
ZFIN-10247: remove dead sysexecs Makefiles and supporting files

cmpich:
## Summary
- The sysexec mechanism (Informix C UDF reading the `execweb` table to invoke server-side scripts) was never ported to PostgreSQL, so the entire `server_apps/sysexecs/` apparatus has been dead since the Postgres migration. A grep across `*.java`/`*.groovy`/`*.pl`/`*.sh`/`*.sql`/`*.xml`/`*.gradle`/`*.tag`/`*.jsp` shows zero callers of `sysexec(`, `execweb`, `get_image_stats`, or `make_thumbnail`.
- Drops the four Makefiles listed in [ZFIN-10247](https://zfin.atlassian.net/browse/ZFIN-10247) plus the only two payload scripts (`get_image_stats.pl`, `make_thumbnail.sh`) — the entire `server_apps/sysexecs/` directory.
- Removes the matching `server_apps;sysexecs;deployFiles` Gradle task (and its entry in the main `make` task list) and the `sysexecs` entry in the legacy `server_apps/Makefile` SUBDIRS line.
- **Adds Liquibase migration `1182/migrations/ZFIN-10247.sql` to drop the `execweb` table.** The 3 zombie rows (`get_image_stats`, `make_thumbnail`, `rm`) pointed at paths that no longer exist after this PR; the table has no live readers. Uses `DROP TABLE IF EXISTS` so it's safe to re-run.

## Test plan
- [ ] `./gradlew make` runs cleanly without the `server_apps;sysexecs;deployFiles` task
- [ ] Spot-check that the deploy target no longer contains `server_apps/sysexecs/`
- [ ] Liquibase changeset `ZFIN-10247-drop-execweb-table` runs cleanly on a fresh DB
- [ ] After migration: `SELECT to_regclass('execweb')` returns NULL
- [ ] No regression in image-handling JSPs/pages (none use `sysexec` — sanity check anyway)

[ZFIN-10247]: https://zfin.atlassian.net/browse/ZFIN-10247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rtaylorzfin (approved):
Looks good! I'm happy to see this unused code go away

PR #1858, opened 2026-05-12
ZFIN-8948: Add HTML report viewer for GAF/GOA load

rtaylorzfin:
Use HTML report view for GAF-GOA load.
Update NCBI Report to use the same view to ensure it can be made compatible with other loads.


rtaylorzfin:
Closing this in favor of #1874

PR #1860, opened 2026-05-13
ZFIN-10287: fix assembly-not-known save (stale FGL + FGMD seqRef)

cmpich:
## Summary

Fixes [ZFIN-10287](https://zfin.atlassian.net/browse/ZFIN-10287). When a curator switches a mutant to *"Assembly information not known as of <date>"* by entering a date and clearing the chromosome/assembly/start/end/evidence fields, the save failed and the previously calculated Sequence of Reference kept showing up.

Two bugs combined to produce the symptom:

1. **Stale `FeatureLocation` after delete.** `FeatureRPCServiceImpl.updateFeatureLocation`'s delete branch removed the DB row but left the in-memory Java object's `chromosome / assembly / startLocation / endLocation` populated. The follow-up `validateReferenceSequence(fgmd, fgl)` then read those stale fields, skipped its early-return guard, and tried to fetch the reference sequence from the FASTA at the soon-to-be-removed coordinates. On prod that throws `"Sequence of Reference does not match the auto-generated sequence from the genome assembly"` whenever the calculated value differs (matches the symptom Amy reported). On local Docker the GRCz11 FASTA is also missing, so it surfaces as a plain `RuntimeException` wrapping `FileNotFoundException` and a generic 500.
2. **Stale `FeatureGenomicMutationDetail.fgmdSeqRef`.** The GUI carries over the previously calculated value when the curator clears coords; nothing on the server clears it either, so even after a successful save the calculated sequence persists in the DB and is shown on next render.

This PR fixes both by:

- Nulling the in-memory `FeatureLocation` fields in the delete branch of `updateFeatureLocation`, so subsequent callers observe the deleted state coherently.
- Clearing `fgmd.fgmdSeqRef` when `locationDeleted` is true in `editFeatureDTO`, before the validate-and-persist step.

## Reproduction

On local Docker:
1. Open `https://localhost:8446/action/curation/ZDB-PUB-200102-5` (a371's pub).
2. Click into feature **a371**, which already has a calculated `fgmd_sequence_of_reference` and a populated `feature_location` row.
3. Enter a date in *Assembly information not known as of*, clear chromosome/assembly/start/end/evidence, save.

Before the fix: red `Failed to create feature: ... 500 The call failed on the server`. After the fix: the location row is deleted, `fgmd_sequence_of_reference` is nulled, `feature.ftr_chr_info_date` is set.

## Test plan
- [ ] Reproduce on local Docker per the steps above — save succeeds, calculated sequence clears.
- [ ] Verify mz19a path still works (it already did; this PR shouldn't regress it).
- [ ] Verify a non-deletion save (just edit a coordinate) still triggers `validateReferenceSequence` normally and rejects a mismatched seqRef.
- [ ] Verify the assembly-changed path still throws its `"Assembly can only be changed to..."` validation when applicable.

[ZFIN-10287]: https://zfin.atlassian.net/browse/ZFIN-10287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/source/org/zfin/gwt/curation/server/FeatureRPCServiceImpl.java
+++ b/source/org/zfin/gwt/curation/server/FeatureRPCServiceImpl.java
@@ -142,6 +142,15 @@ public class FeatureRPCServiceImpl extends RemoteServiceServlet implements Featu
             dto.getFeatureEndLoc() == null) {
             infrastructureRepository.deleteRecordAttribution(fl.getZdbID(), dto.getPublicationZdbID());
             HibernateUtil.currentSession().delete(fl);
+            // Clear the in-memory fields too. The caller may still hold this reference
+            // (e.g. validateReferenceSequence reads fl.getStartLocation() etc.), and
+            // without this it would keep observing the pre-delete coordinates and
+            // attempt a FASTA lookup against a location the curator just removed.
+            fl.setChromosome(null);
+            fl.setAssembly(null);
+            fl.setStartLocation(null);
+            fl.setEndLocation(null);
+            fl.setLocationEvidence(null);
             return;
         }
         // If the assembly is changing, only allow changing to a current assembly (GRCz12tu or GRCz11)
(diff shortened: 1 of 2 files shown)


rtaylorzfin (approved):
Looks good

PR #1861, opened 2026-05-13
ZFIN-10280: report-only Sequence-of-Reference drift check

cmpich:
## Summary

Adds a read-only weekly validation report for [ZFIN-10280](https://zfin.atlassian.net/browse/ZFIN-10280): compares each feature's stored `fgmd_sequence_of_reference` against the sequence computed live from the assembly FASTA at the feature's stored coordinates, and reports any mismatch. No DB writes.

This is the report-only first pass per the discussion. Once we look at the actual mismatch set and compare it to the 185-row report Ryan attached on the ticket, we can decide whether to flip to auto-update or to hand-curate.

## What's in the PR

- `source/org/zfin/infrastructure/ant/CheckSequenceOfReferenceDriftTask.java` — iterates every `FeatureGenomicMutationDetail` with a non-null `fgmdSeqRef`, joins to its `FeatureLocation`, calls `GenomicLocationService.getReferenceSequence(...)`, and records mismatches as report rows. Skips features whose assembly FASTA isn't present in the environment (logs and continues — so e.g. test-with-GRCz11-only doesn't crash).
- `server_apps/DB_maintenance/build.xml` — new `check-sequence-of-reference-drift` Ant target that hands the job over to the standard validate-data report writer.
- `server_apps/DB_maintenance/validatedata/report.properties` — `errorMessage` + `headerColumns` entries for the report template.
- `server_apps/jenkins/jobs/Check-Sequence-Of-Reference-Drift_w/config.xml` — Jenkins job (weekly cadence implied by the `_w` suffix). The trigger `<spec>` is empty for now — turn it on after a sanity-check run on test.

## Report columns

| Feature ID | Feature Abbreviation | Feature Type | Assembly | Chromosome | Start | End | Stored Sequence of Reference | Calculated (FASTA) Sequence |

## Test plan
- [ ] Trigger the job manually on test once a build is up.
- [ ] Confirm the HTML report writes to `server_apps/DB_maintenance/validatedata/Check-Sequence-Of-Reference-Drift_w/`.
- [ ] Cross-check the mismatch count and rows against the existing 185-row report attached to ZFIN-10280.
- [ ] Sanity-check a few rows by hand (one that should mismatch, one that shouldn't) to confirm the comparison and the column values.
- [ ] Once happy, set the Jenkins trigger spec (e.g. `H H * * 0` for weekly) so the report runs unattended.

## Follow-ups (not in this PR)
- Decide between auto-update vs. curator workflow once the mismatch set is reviewed.
- Optional: add the per-feature "Recalculate Sequence of Reference" button Ryan suggested, as a separate small PR.

[ZFIN-10280]: https://zfin.atlassian.net/browse/ZFIN-10280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server_apps/DB_maintenance/build.xml
+++ b/server_apps/DB_maintenance/build.xml
@@ -339,6 +339,21 @@
         </java>
     </target>
 
+    <target name="check-sequence-of-reference-drift" description="ZFIN-10280: report features whose stored fgmd_sequence_of_reference disagrees with the FASTA-computed sequence at the stored coordinates. Read-only.">
+        <echo message="Check Sequence of Reference drift"/>
+        <echo message="Use database: ${DBNAME}"/>
+        <echo message="Use JobName: ${JobName}"/>
+        <echo message="ValidateData Dir: ${env.TARGETROOT}/server_apps/DB_maintenance/validatedata"/>
+        <java classname="org.zfin.infrastructure.ant.CheckSequenceOfReferenceDriftTask" fork="yes"
+              classpathref="combined.classpath"
+              failonerror="true">
+            <arg value="${DBNAME}"/>
+            <arg value="${JobName}"/>
+            <arg value="${env.TARGETROOT}/server_apps/DB_maintenance/validatedata"/>
+            <arg value="${web-inf.dir}/zfin.properties"/>
+        </java>
+    </target>
+
     <target name="run-data-report-param" description="">
         <data-report-param/>
     </target>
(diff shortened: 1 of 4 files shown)


rtaylorzfin (commented):
@cmpich  This looks good to me. I see the ticket is still in the product backlog. Should we move it into current sprint? Also, the ticket mentions comparing deletion size to start and end location. Perhaps we should include that here too? That recently got changed to an automatic calculation, but the DB might have old data.

rtaylorzfin (approved):
Looks good

PR #1863, opened 2026-05-13
ZFIN-10290: propagate construct edit validation errors to curator

rtaylorzfin:
Wrap updateConstruct in try/catch mirroring createAndUpdate so InvalidConstructNameException returns its message in ConstructUpdateResult instead of falling through to the global handler's HTML 500 page, which the frontend could not parse as JSON. Also rolls back the transaction on failure.


cmpich (approved):
Looks good.

PR #1866, opened 2026-05-13
ZFIN-10208: improve daily digest with stale-failure list

rtaylorzfin:
Add a "Failing Jobs (last run failed)" section to the bottom of the Jenkins daily summary email so jobs whose most recent build failed outside the reporting window stay visible until they're fixed. Render a colored-dot sparkline of each job's last four build results next to its name in every table.


cmpich (approved):
Looks good, though, that long script in the XML feels strange. Could that be an separate, external script being called so it does not mix XML with groovy?

PR #1867, opened 2026-05-13
ZFIN-10294 & ZFIN-10295: Add login option to captcha challenge; preserve redirect through login

rtaylorzfin:
Logged-in users bypass the captcha. Give visitors a way out without solving altcha by adding a "Log in" button to the challenge page that carries the original URL through the login flow and back.

To make redirect-after-login work generally, also rewrite the header "Sign In" link client-side to include the current page URL, so signing in from any page returns the user to that page.

The redirect parameter flows: link/button -> /action/login -> login form hidden input -> j_security-check -> success handler. Wire it up by:

  - LoginController: forward redirect when bouncing /login -> /login-redirect
  - login_form.jsp: render hidden redirect input from param
  - security.xml: set targetUrlParameter=redirect on success handler
  - ApgAuthenticationSuccessHandler: override determineTargetUrl to reject anything that isn't a site-relative path or same-host URL, closing the open-redirect vector the targetUrlParameter would otherwise expose
  - MigratingAuthenticationFailureHandler: preserve redirect on failure so a retry still lands on the original page, and stop mutating the shared defaultFailureUrl field (the previous setDefaultFailureUrl pattern raced across concurrent failed logins)

### DIFF
--- a/home/WEB-INF/jsp/infrastructure/captcha-altcha-challenge.jsp
+++ b/home/WEB-INF/jsp/infrastructure/captcha-altcha-challenge.jsp
@@ -19,7 +19,7 @@
                     </p>
 
                     <form id="captcha-form" method="post" action="/action/captcha/challenge">
-                        <input type="hidden" name="redirect" value="${captchaRedirect}"/>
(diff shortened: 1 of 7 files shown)


cmpich (approved):
Looks good.

PR #1868, opened 2026-05-14
ZFIN-10298: hide emails from search and block Community downloads

rtaylorzfin:
Stop the Faceted search from exposing email addresses:
- Refuse the Community CSV download (hide button + 403 the endpoint).
- Drop the Email attribute and column from Person search results and the community results table view.
- Always suppress the Email Address Solr highlight on Person/Lab hits.
- Drop the email cell from the person search results tag.


cmpich (approved):
Looks good.

PR #1872, opened 2026-05-15
ZFIN-10278: search page Feedback button opens Your Input Welcome dialog

rtaylorzfin:
The search page's Feedback button opened a redundant inline form. Route it to the same Your Input Welcome modal triggered from the bottom-right of every page, and drop the now-unused feedbackModal tag and CSS.


cmpich (approved):
Looks good.

PR #1873, opened 2026-05-18
ZFIN-10259 Add monthly UniProt re-run job

rtaylorzfin:
Schedule a new Jenkins job, Uniprot-Monthly-Diff-Load_m, that runs on the 1st of each month and re-runs the UniProt primary and secondary loads against the latest processed release file so DB-side changes get integrated even when no new UniProt release is available. If an unprocessed release exists, the job marks the build UNSTABLE and skips the loads so the manual UniProt-Diff-Load / UniProt-Secondary-Term-Load flow handles the new release.

UniProtLoadTask and UniprotSecondaryTermLoadTask gain a UNIPROT_RERUN_LATEST_PROCESSED env flag that falls back to the latest processed release (and assigns the release entity so the relevant upr_*_date column is refreshed on the re-run).

### DIFF
--- a/server_apps/jenkins/jobs/Uniprot-Monthly-Diff-Load_m/config.xml
+++ b/server_apps/jenkins/jobs/Uniprot-Monthly-Diff-Load_m/config.xml
@@ -1,0 +1,141 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<project>
+  <actions/>
+  <description>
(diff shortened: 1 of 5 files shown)


cmpich (commented):
Looks good.

rtaylorzfin:
Closing this to replace with #1874

PR #1874, opened 2026-05-19
ZFIN-8948 & ZFIN-10259: Uniprot monthly load and Load-GAF-GOA reports

rtaylorzfin:
This replaces the 2 other pull requests because they overlap on the report generation part. This resolves those conflicts.

Run Uniprot load monthly.

Use HTML report view for GAF-GOA load.
Update NCBI Report to use the same view to ensure it can be made compatible with other loads.


source/org/zfin/sequence/load/LoadActionReportAdapter.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private LoadReportMeta buildMeta(String title) {
+        LoadReportMeta meta = new LoadReportMeta();
+        meta.setTitle(title);
+        meta.setCreationDate(System.currentTimeMillis());
+        return meta;
+    }
+
+    private LoadReportSummary buildSummary(EnsemblLoadSummaryItemDTO dto) {
```
cmpich, line 59: Can that be omitted and the new style be called directly?

rtaylorzfin: @cmpich I tried to update this just now to use the new style directly. I ended up just re-implementing these translation tools inline in the load logic. After looking at the results, I think this adapter pattern is actually still cleaner.

cmpich (approved):
Looks good.

PR #1876, opened 2026-05-20
ZFIN-10304: comments, section approval, and status overview bar

cmpich:
- Per-field/per-section curator/submitter comment threads (new zirc.line_submission_comment table, REST endpoints, shared modal opened from a hover-revealed comments icon).
- Per-section Approved checkbox (new zirc.line_submission_section_approval) enabled only when rollup status is Complete; toggling propagates to parent section badges and the overall page status; each toggle is logged to the updates audit table.
- Page-top Status Overview bar that mirrors the top-level sections with Missing/In-Progress/Complete/Approved colors and an inline status legend.
- web.xml: UTF-8 jsp-config so literal non-ASCII characters in JSPs render correctly.
- run_zfin10241_backfill.sh: WAR_LIB env-var override so the script works in environments whose deployed WAR lives outside TARGETROOT.

### DIFF
--- a/source/org/zfin/db/postGmakePostloaddb/1182/run_zfin10241_backfill.sh
+++ b/source/org/zfin/db/postGmakePostloaddb/1182/run_zfin10241_backfill.sh
@@ -49,7 +49,10 @@ hdr "ZFIN-10241 backfill — preflight checks"
 
 PMID_FILE="$SOURCEROOT/$PMID_FILE_REL"
 PUBMED_DIR="$TARGETROOT/server_apps/data_transfer/PUBMED"
-WAR_LIB="$TARGETROOT/home/WEB-INF"
+# WAR_LIB defaults to the schlapp/dev exploded layout under $TARGETROOT;
+# trunk + other environments where the WAR lives under catalina_bases can
+# override with e.g.  WAR_LIB=/opt/zfin/catalina_bases/zfin.org/webapps/ROOT/WEB-INF
+WAR_LIB="${WAR_LIB:-$TARGETROOT/home/WEB-INF}"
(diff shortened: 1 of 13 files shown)


source/org/zfin/db/postGmakePostloaddb/1182/run_zfin10241_backfill.sh
```diff
@@ -49,7 +49,10 @@ hdr "ZFIN-10241 backfill — preflight checks"
 
 PMID_FILE="$SOURCEROOT/$PMID_FILE_REL"
 PUBMED_DIR="$TARGETROOT/server_apps/data_transfer/PUBMED"
-WAR_LIB="$TARGETROOT/home/WEB-INF"
```
rtaylorzfin, line 52: Does this fail without setting CP? The GROOVY_CLASSPATH variable should resolve to those directories. Maybe we need to fix GROOVY_CLASSPATH if it's not working

rtaylorzfin (approved):
This is looking good. I think the caveat on our work on all the ZIRC Submission PRs is that everything is a work-in-progress so we may need to make changes later. The commenting on sections is interesting. We end up with sections having an implied relationship to fields, but I think that's fine.  I wonder as we continue to grow the detail page and the edit page, if we'll reach a point eventually where it makes sense to combine them in some way. I could imagine at least both of them reading from the same schema. 

I think we could do some more iterations on the design. I'm not great at design myself, but all of the icons everywhere indicating status starts to feel a little noisy or cluttered (IMO).

I'll also request a Co-Pilot code review to see if it turns up anything interesting.

Thanks!

PR #1882, opened 2026-05-27
ZFIN-10278: skip feedback-form captcha for logged-in / already-verified users

rtaylorzfin:
The Your Input Welcome feedback form always showed an Altcha captcha. Omit it when a captcha is not actually required, reusing the rules in getRedirectUrlIfNeeded (logged in, flag off, bypassed IP, or already verified).

- CaptchaService: extract isCaptchaRequired(request); getRedirectUrlIfNeeded now delegates to it.
- CaptchaController: add GET /captcha/required returning {"required": bool} so the frontend can decide whether to present the widget.
- UserCommentController: only enforce captcha when isCaptchaRequired; return a distinct "CaptchaRequired" status (not a generic error) so the client can recover.
- your-input-welcome.js: consult the endpoint on open; hide the widget and enable submit when not required; on a CaptchaRequired submit response (e.g. session expired), reveal the widget and let the user resend without losing their comment.
- yourInputWelcome.tag: captcha hidden by default; JS reveals it only when needed; statechange handling moved into the JS file.

### DIFF
--- a/home/WEB-INF/tags/yourInputWelcome.tag
+++ b/home/WEB-INF/tags/yourInputWelcome.tag
@@ -41,7 +41,12 @@
                         <label for="input-welcome-comments">Comments:</label>
                         <textarea id="input-welcome-comments" name="yiw-comments"></textarea>
                     </div>
-                    <div class="control" style="padding-left: 135px; height: 80px;">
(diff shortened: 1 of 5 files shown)


cmpich (approved):
Looks good.

PR #1886, opened 2026-05-28
ZFIN-10316: trim whitespace in NumberTextBox coordinate fields

rtaylorzfin:
Pasting trailing/leading whitespace after a coordinate (e.g. "246810 ") caused StringUtils.isNumeric to return false, so getBoxValue returned null. That blocked reference-sequence auto-generation and produced a validation error when adding the line.

Trim the text before validating/parsing in getBoxValue, isValid, and isFieldEqual so surrounding whitespace is ignored across all numeric curation inputs.


cmpich (approved):
Looks good.

PR #665, opened 2023-05-31
ZFIN-8643 ZFIN-8671 ZFIN-8653 make figure and image prototype pages public pages, modify citation to citations

cmpich:

PR #1887, opened 2026-05-28
ZFIN-10314: carry probe through figure page links and show Probe

rtaylorzfin:
For Thisse / direct-submission figures, the figure page is viewed in the context of a single probe. Restore probe-awareness that was lost in the figure page rewrite (#665):

- "Back to All Figure Page" link now passes ?probeZdbID so it lands on the probe-filtered all-figures page (< 50 figures) instead of the publication-figures-unavailable dead end.
- "Other Figures" list is now probe-limited via getFiguresForPublicationAndProbe rather than the full publication's figures.
- Add a "Probe" row to the figure summary linking to the clone.


cmpich (approved):
Looks good.

PR #1892, opened 2026-06-01
ZFIN-10314: Figure link improvement

rtaylorzfin:
The image page should handle large data publications in the same way as the figure page:

Hide "all figures" link for pubs with a lot of figures, but show the link if the pub has probe information and limit the results to only the probe figures.


cmpich (approved):
Looks good.

PR #1893, opened 2026-06-01
ZFIN-10277: make disease/phenotype page ordering deterministic

rtaylorzfin:
The order-by clauses in HibernateDiseasePageRepository only sorted on non-unique display columns (fish order/name, gene symbol, display name). Rows tying on those columns had undefined relative order, so PostgreSQL could return them differently between executions, producing unstable ordering and inconsistent pagination across pages.

Append each row's natural key (persistent ZDB IDs, stable across table rebuilds) as a tie-breaker to establish a total order on all four queries: getPhenotype, getFishDiseaseModels, getFishDiseaseChebiModels, and getPhenotypeChebi. Display ordering is unchanged for non-tied rows.

Test endpoints (results should be identical across repeated requests and consistent page-to-page):

  getPhenotype (the reported case):
    /action/api/ontology/ZDB-TERM-100331-504/phenotype?directAnnotation=true&limit=25&page=1
    /action/api/ontology/ZDB-TERM-100331-504/phenotype?directAnnotation=false&limit=25&page=1

  getPhenotypeChebi:
    /action/api/ontology/ZDB-TERM-100331-504/phenotype-chebi?limit=25&page=1

  getFishDiseaseModels:
    /action/api/ontology/ZDB-TERM-100331-504/zebrafish-models?limit=25&page=1

  getFishDiseaseChebiModels:
    /action/api/ontology/ZDB-TERM-100331-504/chebi-zebrafish-models?limit=25&page=1

### DIFF
--- a/source/org/zfin/ui/repository/HibernateDiseasePageRepository.java
+++ b/source/org/zfin/ui/repository/HibernateDiseasePageRepository.java
@@ -105,7 +105,8 @@ public class HibernateDiseasePageRepository implements DiseasePageRepository {
                 hql += "LOWER(" + entry.getKey() + ") like '%" + entry.getValue().toLowerCase() + "%' ";
             }
         }
-        hql += "order by fishStat.fish.order, fishStat.fish.nameOrder, fishStat.geneSymbolSearch";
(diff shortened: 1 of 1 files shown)


cmpich (approved):
Looks good.

PR #1894, opened 2026-06-01
ZFIN-8948 follow-up: add GOA load impact tables to report

rtaylorzfin:
GoaLoadSnapshot captures four HQL counts of marker_go_term_evidence rows attributed to a GAF organization: per evidence code, distinct GO terms, distinct markers, and per source publication. GafLoadJob snapshots the DB once just before the first batch mutation and once after the last; non-GOA loads and snapshot failures are no-ops. GafReportBuilder appends three new tables (Totals, per evidence code, per source pub) inline on the root node just above the Source links, with Before / After / Net change / % change columns. Per-source rows with zero net change are dropped.

Also: declare an explicit dependsOn copyClasspathTemplates on the loadGafGoa gradle task — Gradle 8.14 strict validation flags the task otherwise, since the global wire in root build.gradle only covers root-project JavaExec tasks.

### DIFF
--- a/server_apps/DB_maintenance/build.gradle
+++ b/server_apps/DB_maintenance/build.gradle
@@ -40,6 +40,7 @@ task validateBlastDatabases(type: JavaExec) {
 }
 
 task loadGafGoa(type: JavaExec) {
+    dependsOn ":copyClasspathTemplates"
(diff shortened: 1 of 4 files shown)


cmpich (approved):
Looks good.

PR #1895, opened 2026-06-01
ZFIN-10306: attribute feature-gene attribution history to a real person

rtaylorzfin:
CheckFeatureGeneAttributionTask runs without a logged-in user, so when it was run with COMMIT_CHANGES=true during the release 1176 post-load (2025-12-16) every history row it wrote was attributed to the anonymous "Guest" fallback (submitter_id NULL, submitter_name 'Guest'), producing 8456 'record attribution' rows with submitter 'Guest'.

Require a SUBMITTER_PERSON_ID env var whenever COMMIT_CHANGES is set, resolve it to a Person, and write history via the Person overload so committed changes carry a real submitter. Fail fast (rather than silently logging 'Guest') when the var is missing/blank or names no person.

New invocation:
  COMMIT_CHANGES=true SUBMITTER_PERSON_ID=ZDB-PERS-210917-1 gradle checkFeatureGeneAttributionTask

Add a 1183 Liquibase migration to re-attribute the existing 8456 Guest rows to Ryan Taylor (ZDB-PERS-210917-1, who ran the task) and append " (See ZFIN-9940)" to their comments. Targets exactly those rows (submitter 'Guest' + 'record attribution' + exact comment) and is idempotent.

### DIFF
--- a/source/org/zfin/db/postGmakePostloaddb/1183/migrations/0020-ZFIN-9940-fix-guest-attribution-submitter.sql
+++ b/source/org/zfin/db/postGmakePostloaddb/1183/migrations/0020-ZFIN-9940-fix-guest-attribution-submitter.sql
@@ -1,0 +1,15 @@
+--liquibase formatted sql
+--changeset rtaylor:ZFIN-9940-fix-guest-attribution-submitter
+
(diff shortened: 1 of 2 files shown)


cmpich (approved):
Looks good.

PR #1896, opened 2026-06-02
ZFIN-10286: add persistent orthology-inconsistency report + one-time TSV

rtaylorzfin:
reportOrthoNameChanges.pl now also writes
orthoInconsistentZebrafishGeneNamesReport_persistent.txt at the end of its run. The new sub re-applies the existing inconsistency rule across every zebrafish gene but skips the alreadyExamined allow-list, so curators see the full surface every week. First-seen dates are tracked in /research/zarchive/load_files/Orthology/alreadyExamined.json (loaded at start, fully rewritten at end). Resolved entries drop from the JSON; new entries get today's date; existing entries keep their original date. Report is sorted newest-first with the first-seen date above each block. Filename matches the existing orth*.txt Jenkins attachment glob.

oneTimeOrthoInconsistencyReport.pl is a standalone scan of Human + Mouse orthologs as stored in the ortholog table — same match rule, no allow-list, TSV output. PG* env vars override the deployed placeholder defaults so it runs anywhere.


rtaylorzfin:
Closing in favor of PR #1904

PR #1897, opened 2026-06-02
ZFIN-10278 follow-up: novalidate the feedback form so hidden Altcha checkbox stops blocking submit

rtaylorzfin:
The Altcha widget injects a hidden <input type="checkbox" required>. When the server tells us the visitor doesn't need a captcha (logged-in or already verified), #input-welcome-captcha is hidden via display:none — but the required checkbox is still in the DOM, so the browser's native form validation fires on submit and logs "An invalid form control with name='' is not focusable."

The form's JS does its own field validation (see `validators` in your-input-welcome.js) and gates the submit button on captcha state, so we don't need the browser's pass. novalidate disables it cleanly and is robust to Altcha widget internals.

### DIFF
--- a/home/WEB-INF/tags/yourInputWelcome.tag
+++ b/home/WEB-INF/tags/yourInputWelcome.tag
@@ -8,7 +8,15 @@
             Your Input Welcome
         </div>
         <div class="popup-body" id="input-welcome-body">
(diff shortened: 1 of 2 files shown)


cmpich (approved):
Looks good.

PR #1903, opened 2026-06-04
ZFIN-10327: Remove Makefiles

rtaylorzfin:
This pull request removes 16 obsolete Makefiles from the `server_apps/data_transfer/` directory. These Makefiles have been non-functional since the removal of the shared `make.include` file.

* Added `MAKEFILE_AUDIT.md` to list the remaining Makefiles that require further review before removal. The document also identifies potentially orphaned files and proposes a follow-up ticket for the remaining Makefiles.


cmpich (approved):
Looks good.

PR #1789, opened 2026-03-28
ZFIN-10216: Slimmer docker images possible improvements

rtaylorzfin:
This pull request removes the dependency on ImageMagick's `convert` binary for image thumbnail and medium image generation, replacing it with a Java-based solution implemented in `ImageService`. It also cleans up related configuration, scripts, and Docker setup, and introduces a local build cache for Docker dependencies.

The most important changes are:

**Removal of ImageMagick and related scripts/configs:**
**Switch to Java-based image processing:**

* Updated the thumbnail/medium image generation logic in `getPDFandImages.groovy` to use `ImageService` methods instead of shelling out to `convert`
* Refactored and expanded Java imports in `ImageService.java` to support direct image manipulation.

**Docker build improvements and dependency caching:**

* Refactored `docker/base/Dockerfile` to streamline package installation, remove ImageMagick, and introduce a local cache for build dependencies (Bowtie, Groovy, etc.) using a new `deps.sha256` manifest.
* Updated `docker/docker-compose.yml` and environment files to support specifying a local Docker build cache path for dependencies since downloading from sourceforge and others can be slow

**Cleanup of obsolete files:**

* Removed the outdated `apache-groovy-binary-4.0.29.zip.sha256` file and replaced its functionality with the new `deps.sha256` 

### Even more potential improvements
Here is where a lot of our space is going in this image:
```
  ┌───────┬───────────────────────────────────────────────────────────┐
  │ Size  │                           What                            │
  ├───────┼───────────────────────────────────────────────────────────┤
  │ 193MB │ Base JDK 21 image (Eclipse Temurin)                       │
  ├───────┼───────────────────────────────────────────────────────────┤
  │ 152MB │ Gradle 8.14.4 download + install                          │
  ├───────┼───────────────────────────────────────────────────────────┤
  │ 516MB │ apt install (PostgreSQL client, Node.js, Perl, ant, etc.) │
  ├───────┼───────────────────────────────────────────────────────────┤
  │ 555MB │ Bowtie + Bowtie2 + Groovy (extracted)                     │
  └───────┴───────────────────────────────────────────────────────────┘
```

### Bowtie takes a lot of space! 
This makes me wonder if there is something we can do to improve the bowtie handling. Perhaps it could reside in a volume or something like that?

### Break down of apt packages
The other big chunk is the stuff that we install with apt. We could take a closer look at some of that:
```
┌───────┬─────────────────────────┬─────────────────────────────────────────────────────┐
[...]

PR #1905, opened 2026-06-06
ZFIN-10216: Slimmer docker images (de-BuildKit, bowtie via mount)

rtaylorzfin:
Re-applies PR #1789 on current main, slimming the base image from ~4.0GB to ~2.2GB, with these adaptations:

- No BuildKit: base image uses no BuildKit-only syntax (NodeSource key is an ASCII .asc in trusted.gpg.d, no gpg --dearmor/--mount). Replaced the ADD --checksum postgres JDBC fetch in tomcat and solr Dockerfiles with curl + sha256sum (run as root); each keeps a comment showing the ADD --checksum one-liner to restore if BuildKit is re-enabled. Hardcoded "noble" in compile/Dockerfile since lsb-release is dropped.
- Bowtie/bowtie2 mounted from the host like ab-blast (DOCKER_BOWTIE_PATH -> /opt/misc/bowtie, /opt/misc/bowtie2) instead of baked in (~415MB) or shimmed.

Base image apt layer consolidated and cache cleaned in-layer; dropped imagemagick, emacs, build-essential, cpanminus, lsb-release; purged unused inherited packages (python3-breezy, mercurial, subversion, binutils, git-lfs). Perl JSON kept via apt libjson-perl (no build-essential needed).

ImageMagick replaced with pure-Java image processing: ImageService now uses ImageIO (aspect-preserving resize, CMYK->RGB, multi-reader fallback) instead of shelling out to convert; getPDFandImages.groovy calls ImageService. Removed CONVERT_BINARY_PATH property and the dead fix_medium_thumb.sh.


cmpich (approved):
Looks great!

PR #1906, opened 2026-06-08
ZFIN-10325 add submitters/PIs support

cmpich:


rtaylorzfin (approved):
Looks good. I wonder if we should dedicate a ticket to adding some tests around some of these new submission features. Nice work!

PR #1907, opened 2026-06-08
ZFIN-10207: Port marker merge to Java and retire the Perl CGI

rtaylorzfin:
Replace cgi-bin/merge_markers.pl with a Java implementation shared by the webapp and a new zfin-util command, proven byte-for-byte equivalent to the Perl before the Perl was removed.

- ToolBootstrap: headless entry point so command-line jobs run with ZfinProperties + Hibernate (one managed transaction, no Spring context). CommandLineUtilityRegistry gains register(..., requiresDatabase=true).
- MarkerMergeService: faithful port of the Perl's information_schema foreign-key walk plus the marker business rules; runs on the caller's Hibernate connection without owning the transaction. Exposed via `zfin-util merge-markers <delete> <into> [--dry-run] [--skip-regen]`.
- Webapp: MergeMarkerController POST /action/marker/merge now uses the service and redirects to the survivor; merge-marker.jsp / merge-antibody.jsp POST there instead of the cgi. The webapp now enforces same-type merges (stricter than the old cgi).
- Verified equivalent across 100 pairs / 32 tables by a differential harness; retired the Perl (browser-verified gene/antibody/error paths on cell-mac). Deleted cgi-bin/merge_markers.pl, the whole server_apps/DB_maintenance/merge/ dir (harness, cmdline variant, SQL-gen workflow) and the now-empty cgi-bin deploy plumbing; Perl + harness archived to the retirement ticket. Removed httpd's perl/cgi support (perl-* packages, the perl symlink, and the /cgi-bin ScriptAlias / Location / Directory config).
- Build: register home/WEB-INF/classes as a main source-set output (builtBy: copyClasspathTemplates) so tasks gain the dependency automatically and stop tripping Gradle 8.x strict dependency validation.


cmpich (approved):
I assume you have done good testing of the converted code?

rtaylorzfin:
> I assume you have done good testing of the converted code?

@cmpich Yes, I did. I automated some testing that merged 100 pairs of genes using the old perl code and then did the same thing with the java code and compared the results. I made sure that the results were identical. Even found some things that were possibly bugs in the perl code. I created a follow-up ticket to see if we want to update that behavior.

PR #1911, opened 2026-06-10
HTML Report improvement: compress internal data

rtaylorzfin:
Our HTML reports get pretty large sometimes. We can compress them as zip files but that can be annoying to go through the trouble of downloading and unzipping each time. This uses gzip (supported by Java and Javascript natively) to compress the JSON and encode it as base64 inside the html--making the files much smaller.


cmpich (approved):
Looks great!

PR #1912, opened 2026-06-10
ZFIN-9394: Fix SignaFish load: https source, empty-input guard, per-ID link adds

rtaylorzfin:
The Load-Signafish_w job wiped all 177 SignaFish db_links (build #516). Root cause: the source URL moved to https, but parseFile() fetched http://signalink.org/zfin_ids.lst. URL.openStream() does not follow cross-protocol (http->https) redirects, so it read the empty 302 body -> 0 IDs -> deleteMarkerDBLinksNotInList removed every existing link.

Changes:
- LoadSignafishJob.parseFile(): use https; trim/skip blank lines; log the URL and the line/ID counts read.
- LoadSignafishJob.load(): abort (rollback, fail the build) if the parsed list is empty rather than silently deleting every link; guard against a null reference-database lookup; log before/after counts to the console and report.
- HibernateMarkerRepository.addMarkerDBLinks(): add links per-ID instead of all-or-nothing. It previously skipped the entire batch if ANY link already existed, so newly added source IDs were never linked in steady state (e.g. gdf6a / ZDB-GENE-980526-373). Now only missing IDs are added, unknown markers are skipped, and the count added is returned.

Verified against a local DB: fetches 178 IDs over https, links the previously-orphaned gdf6a (177 -> 178), and a repeat run is a clean no-op with no duplicates.

### DIFF
--- a/source/org/zfin/datatransfer/LoadSignafishJob.java
+++ b/source/org/zfin/datatransfer/LoadSignafishJob.java
@@ -2,12 +2,9 @@ package org.zfin.datatransfer;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.zfin.Species;
 import org.zfin.framework.HibernateUtil;
 import org.zfin.infrastructure.ant.AbstractValidateDataReportTask;
-import org.zfin.repository.RepositoryFactory;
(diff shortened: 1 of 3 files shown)


cmpich (approved):
Looks good.

PR #1916, opened 2026-06-24
ZFIN-10357: Fix PMC "Original Article" PDF classification

rtaylorzfin:
## Problem

When the PMC/EuropePMC document-acquisition job (`getPDFandImages.groovy`) moved off FTP to the Open Access S3 API, the "Original Article" marking broke. The job:
- picked an arbitrary PDF as the main article (`s3Files.find { endsWith('.pdf') }` lands on a supplement, since S3 keys sort lexicographically — e.g. `NIHMS…-supplement-10.pdf` before `PMC….pdf`), and

## Ingest fix (`getPDFandImages.groovy`)

- `classifyPdfs()`: identify the main article PDF as the one whose basename matches the package `.xml`/`.nxml` basename (the accession) — the same robust signal the Alliance literature pipeline uses. All other PDFs are supplements.
- Record only the main as Original Article; route the rest through `recordSupplementFromS3()` → Supplemental Material. Falls back to the first PDF (with a logged warning) only when no XML match exists, so an article is never silently dropped.

## Historical cleanup

One-time repair for rows already mislabeled. Auto-fixes only the unambiguous cases:
- **A**: duplicate rows pointing at the same physical file → delete the redundant row.
- **B**: one genuine main + clearly-named supplements all marked type 1 → demote the supplements (only when exactly one main remains, so the sole article is never demoted).

### DIFF
--- a/home/WEB-INF/jsp/publication/publication-view-summary.jsp
+++ b/home/WEB-INF/jsp/publication/publication-view-summary.jsp
@@ -103,7 +103,7 @@
             <c:forEach items="${publication.files}" var="file" varStatus="loop">
                 <a href="${ZfinPropertiesEnum.PDF_LOAD.value()}/${file.fileName}">
                     <c:choose>
-                        <c:when test="${file.type.name == 'ORIGINAL_ARTICLE'}"><b>${file.originalFileName}</b></c:when>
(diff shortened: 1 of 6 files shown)


cmpich (approved):
looks good.

PR #1917, opened 2026-06-24
ZFIN-10337: Gene updates page should include changes for its old ID 

rtaylorzfin:
Audit-history rows in the updates table are not repointed when a marker is merged (both the legacy Perl and the Java MarkerMergeService intentionally skip the updates table). Resolve this at the display layer instead of mutating the audit table at merge time.

- Add reverse lookup InfrastructureRepository.getReplacementsForNewZdbID to find the IDs merged into a surviving marker via zdb_replaced_data.
- UpdatesService now folds in the update history of every merged-away ID (transitively, cycle-safe) alongside the survivor's own history.
- Expose recordID on UpdatesDTO and add a "Record" column to the updates table indicating which marker each history row is/was attached to (merged rows are labeled).

Follow-up from ZFIN-10207 (port merge_markers.pl to Java).


cmpich (approved):
Looks good.

PR #1918, opened 2026-06-25
ZFIN-10350: cascade pheno_term_fast_search foreign key deletes

rtaylorzfin:
- pheno_term_regen.sql: re-add the FK with ON DELETE CASCADE every rebuild, and flag (TODO) that this table could itself go incremental to drop the mismatch.
- 0060 migration: convert the FK already on the live table once at deploy, since the failing apply step runs before pheno_term_regen re-creates it.


cmpich (approved):
Looks good.

PR #1919, opened 2026-06-25
ZFIN-10286: soften diff highlight + add gene-symbol columns to report

rtaylorzfin:
Feedback on the orthology-inconsistency report:

- Tone down the diff-highlight color from #ffeb3b (vivid Material yellow) to a paler #fff3a3 — less aggressive on screen while still clearly yellow and still preserving the <u> underline Excel relies on for export.

- Add gene-abbreviation columns to the weekly Update-Orthology_w report's inconsistency tables. The human/mouse symbols were already parsed from the Perl output (hSym/mSym) but never rendered; now both tables show ZFIN / Human / Mouse symbol columns alongside the name columns, character-diff highlighted the same way, so abbreviation-only changes stand out. The one-time OrthoInconsistencyReportJob already had both symbol columns.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,13 @@ docker/elasticsearch/state/
 .idea
 *.iml
 
+# Eclipse things
+.classpath
+.project
+.factorypath
+.settings/
+bin/
+
 # Mac cruft
 .DS_Store
 ._*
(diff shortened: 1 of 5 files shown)


cmpich (approved):
Looks good.

PR #1921, opened 2026-06-29
ZFIN-10276: Fix Solr upgrade ClassCastException: Long facet counts cast to Integer

rtaylorzfin:
The Solr upgrade now returns Long (not Integer) for JSON facet count/num_terms/numBuckets values. Cast to Number and call intValue() so both Integer and Long responses work.


zfinryanm (approved):
Looks good.

PR #1925, opened 2026-07-02
ZFIN-10080 lift over features from GRCz11 to GRCz12tu

cmpich:


rtaylorzfin:
Looks good as far as I can tell

rtaylorzfin (approved):

PR #1927, opened 2026-07-06
ZFIN-10372 Lift over 'la' insertion features Zv9 -> GRCz12tu

cmpich:
## ZFIN-10372 — Lift over 'la' insertion features Zv9 → GRCz12tu

Adds Liquibase migration `1184/migrations/0030-la-insertion-grcz12tu-liftover.sql` establishing GRCz12tu genomic locations for **3742** 'la' transgenic-insertion features that previously had a position only on Zv9.

- Coordinates produced by UCSC liftOver through the assembly chain (Zv9 → GRCz10 → GRCz11 → GRCz12tu).
- Restricted to the **gene-validated** set: each insertion is 'is allele of' a single gene, and its lifted position keeps the same relative position within that gene span (≤5% shift, or under an assembly inversion/strand flip).
- Each new `sfcl` row: `sfcl_zdb_id` minted via `get_id('SFCL')`, registered in `zdb_active_data` before the insert (FK), `sfcl_evidence_code = ZDB-TERM-170419-312` ('evidence used in automatic assertion'), and the GRCz12tu RefSeq accession from liftOver.
- `NOT EXISTS` guard keeps it idempotent — features that already have a GRCz12tu row are skipped.

Independent of the ZFIN-10080 welltested-liftover changesets (0010/0020) in the same directory; safe to apply on its own.


rtaylorzfin (approved):
I assume these are correct. Looks good. Nice work!

PR #1928, opened 2026-07-06
ZFIN-10371: Override auto-calculated fields in feature edit

rtaylorzfin:
If a feature does not have assembly information, allow the otherwise auto-calculated fields to be entered manually:
- Sequence of Reference
- deletion size


cmpich (approved):
looks good.

PR #1929, opened 2026-07-06
ZFIN-10373: Solr backup/restore: per-instance backup, restore from shared zfindb

rtaylorzfin:
Change the Solr snapshot layout so an instance backs up into its own subdir and restores from the shared production snapshot, instead of both sharing one flat dir under the mount.

- backup (solr-ant.xml backup-solr-index): write <base>/$INSTANCE/v9 (overridable via -Dsolr.unloads.dir) so a stage/dev index build never clobbers the snapshot everyone restores from.
- restore (build.gradle getLatestSolrIndex): read <base>/zfindb/v9 by default; override with -PsolrSourceInstance=<name>. zfindb is a symlink to the current prod instance (franklin).
- fetch (build.gradle getLatestSolrUnload): land the scp'd snapshot in <base>/$sourceInstance/v9 to match where restore looks.
- DOCKER_SOLR_UNLOADS_PATH now mounts the parent dir (drop trailing /zfindb) so all per-instance subdirs are visible; the mount point and solr.allowPaths are unchanged.
- docs: build-and-docker.md, solr-reindex.md.


zfinryanm (approved):
Looks good.

PR #1933, opened 2026-07-10
ZFIN-10390: Fix UnknownPathException on disease page fish models

rtaylorzfin:
The order-by clauses in getFishDiseaseModels and getFishDiseaseChebiModels referenced fishModelDisplay.publication, which is not a persistent attribute of FishModelDisplay (publication is only a transient getter over the publications set). Use the mapped singlePublication field (zmd_pub_zdb_id) instead, which Hibernate can resolve.

Introduced in ZFIN-10277 (#1893).


cmpich (approved):
Looks good.

PR #1937, opened 2026-07-14
ZFIN-10374 & ZFIN-10375 & ZFIN-10376 & ZFIN-10377 & ZFIN-10378: refine ZIRC lesion edit form

rtaylorzfin:
- Point mutations: lock Lesion Size read-only at 1bp (DISABLE rule + disabledValue option; renderer displays/persists the fixed value). Enforce lesionSizeBp=1 server-side in updateLesionField so the stored value stays correct regardless of the read-only client field.
- Move Additional Info to the bottom of the lesion fields (all types).
- Capitalize lesion-type display labels (values unchanged).
- Nucleotide change: closed dropdown of the 12 substitutions for point mutations, replacing free text.
- Remove the Location (inline) box (column/DTO retained; whitelisted in the invariants test).
- Rename 5'/3' flank labels to "flanking sequence".
- Regenerate lesion form-schema snapshot.


test/resources/zirc/snapshot/lesion.form-schema.json
```diff
@@ -77,31 +72,48 @@
...
         "type" : "Control"
-      }, {
+      } ],
+      "type" : "Group"
+    }, {
+      "elements" : [ {
         "options" : {
-          "multi" : true
+          "noOther" : true,
+          "standardValues" : [ "A>T", "A>C", "A>G", "T>A", "T>C", "T>G", "C>A", "C>T", "C>G", "G>A", "G>T", "G>C" ],
```
cmpich, line 87: I wonder if that list could go iinto a util file for possible re-use.

cmpich (approved):

test/resources/zirc/snapshot/lesion.form-schema.json
rtaylorzfin: That's a good idea. Could also be generated in code with something like:
```
letters = ['a', 't', 'c', 'g'];
result = letters.flatMap(i => letters.map(j => i + ">" + j));
```

PR #1938, opened 2026-07-17
ZFIN-10382: Static file refactor

rtaylorzfin:
Clean up large static files to make our repo leaner in the following ways:

- Decouple static content from the main application.
  - Moved static content to a separate repository
  - Moved path of static resources to a different volume (/opt/zfin/static)
  - Moved other resources out of ./home (from repository root). It was previously a mix of WEB-INF, front-end assets, static files, and templates
- Pull jenkins plugins using curl instead of keeping them in repository
- Added optional navigation to "zfin book" pages and other similar assets
- Found potential issues for pages that exist but are not linked to


cmpich (approved):
Great jobs!

PR #1939, opened 2026-07-20
ZFIN-10405 set job.output.basedir on ontology-load Jenkins jobs so re…

cmpich:
…port artifacts are archived

buildFile was pointed at $SOURCEROOT (dcbca0073b, for the classpath), which moved ant's basedir and the report output under SOURCEROOT while the Jenkins workspace-relative artifact/HTML/email patterns look under $TARGETROOT. Pass -Djob.output.basedir=$TARGETROOT/server_apps/data_transfer/LoadOntology so the per-job report directory lands in the workspace, matching Load-Zeco-Taxonomy-Ontology.


zfinryanm (approved):
Looks fine to me.

PR #1941, opened 2026-07-23
ZFIN-10202: Adding DOCKER_ARCH env variable to enable arm64 images.

zfinryanm:
Added a DOCKER_ARCH environment variable. Add 'DOCKER_ARCH='-arm64' to your .env for an instance on Mac to use. (Note: The '-' is required to make it work because of limitations in Docker labeling/docker compose.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -857,6 +857,19 @@ task pushLatestGHCRImages() {
     }
 }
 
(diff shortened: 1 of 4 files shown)


rtaylorzfin (approved):
Great! Looking forward to this

PR #1944, opened 2026-07-30
ZFIN-10435: Fix three 500s on term, publication, and phenotype pages

rtaylorzfin:
Found by grouping production 500s in Kibana by tomcat.uri: 5,315 events over two days resolved to a handful of endpoints, 87% of them the first bug below.

1. /action/api/ontology/<term>/chebi-zebrafish-models -- every request, 4,601 hits

     https://zfin.org/action/api/ontology/ZDB-TERM-160831-137635/chebi-zebrafish-models
     https://zfin.org/action/api/ontology/ZDB-TERM-160831-150272/chebi-zebrafish-models

   UnknownPathException: the order-by referenced chebiDisplay.term, but ChebiFishModelDisplay maps the CHEBI term as `chebi` (omca_term_zdb_id). Fails at query-parse time, so it broke every term regardless of data. Introduced in ZFIN-10277 (#1893); the follow-up #1933 fixed `publication` -> `singlePublication` on the same line but left `term`.

2. /action/publication/<pub>/orthology-list -- every request, 478 hits over 293 pubs

     https://zfin.org/action/publication/ZDB-PUB-050823-6/orthology-list
     https://zfin.org/action/publication/ZDB-PUB-071129-8/orthology-list

   "You can't operate on a closed ResultSet" -- getOrthologPaginationByPub combined a tuple transformer with query.scroll(), which Hibernate 6 does not support. HibernateUpgradeHelper.setTupleResultAndListTransformer already documents the constraint; this call site was missed in the upgrade. Use list() like the sibling getOrthologListByPub and paginate the materialized list -- an ortholog list for one publication is small.

3. /action/api/ontology/<term>/phenotype -- the 11 widest terms, 15 hits

     https://zfin.org/action/api/ontology/ZDB-TERM-100331-1056/phenotype
     https://zfin.org/action/api/ontology/ZDB-TERM-100331-35/phenotype

   "PreparedStatement can have at most 65,535 parameters ... Given query has 86,604 parameters". The five "including substructures" queries fetched every matching row id and bound them one-per-parameter via `id in :ids`; the widest terms match 86,664 / 86,604 / 82,104 rows. Replace the id round-trip with a descendant-term subquery over TransitiveClosure (all_term_contains), which binds exactly ONE parameter. all_term_contains carries the self row, so semantics are unchanged -- verified per table that the subquery returns counts identical to the ancestor-array predicate it replaces (82,104 = 82,104 for the widest term; 150/61/332/1 for the rest). findIdsByAncestor is now unused.

   That exposed a second wall underneath: `could not resize shared memory segment ... No space left on device`, because Postgres puts parallel-query DSM segments in /dev/shm and the wide plan climbed through 24 segments to 772MB against a 1g shm_size. shm_size and work_mem must be chosen together -- measured peak /dev/shm and runtime on that query:

     work_mem   1 query           2 concurrent
      128MB       35MB,  8.3s       66MB,  6.7s
      192MB     1027MB,  2.7s           --        needs >2g
      256MB     1027MB,  2.2s     2049MB,   err   fastest, but needs ~4g

   shm_size 2g with work_mem 128MB. 256MB picks a faster parallel-hash plan but cannot fit two concurrent wide queries in 2g, and these pages get crawled. Throttling max_parallel_workers_per_gather to 1 keeps 256MB under 2g but measured slower (9.3s). maintenance_work_mem stays at 1GB, so VACUUM / CREATE INDEX are unaffected.

Verified on a feature stack: all three endpoints return data where they returned 500s (chebi models match the table exactly -- ethanol 61/61, pentetrazol 53/53, metronidazole 38/38; phenotype all-terms counts 74,273 / 78,597 / 78,653), and unrelated terms return totals identical to before (150/80/150, 332/332/332, 179/179/179). orthology-list verified across deep and out-of-range pages.

New DiseasePageRepositoryTest covers all five DiseasePageRepository read methods in three shapes each -- direct, include-children, and with a populated filter map -- plus a widest-term case for the parameter ceiling. Nothing covered this repository before, which is how both an unmapped order-by path and the bind-count ceiling reached production. Reverting either fix fails it.

Known limitation: the widest terms still take ~23-28s, because PaginationResult counts by scrolling all 124k rows to fill a 25-row page. That needs a SQL COUNT plus an id-paged fetch.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -78,7 +78,8 @@ services:
     # configuration that contributed to repeated container OOMs during heavy
     # DIH workloads on a 32g host. Dial back up if you have a clearly RAM-rich
     # host and a workload that benefits.
-    command: -c 'max_connections=7500' -c 'shared_buffers=4GB' -c 'huge_pages=try' -c 'temp_buffers=256MB' -c 'work_mem=256MB' -c 'maintenance_work_mem=1GB' -c 'max_wal_size=10GB' -c 'timezone=America/Los_Angeles' -c 'shared_preload_libraries="pg_stat_statements"'
+    # work_mem is 128MB, paired with the 2g shm_size below -- the two have to be chosen together.
+    command: -c 'max_connections=7500' -c 'shared_buffers=4GB' -c 'huge_pages=try' -c 'temp_buffers=256MB' -c 'work_mem=128MB' -c 'maintenance_work_mem=1GB' -c 'max_wal_size=10GB' -c 'timezone=America/Los_Angeles' -c 'shared_preload_libraries="pg_stat_statements"'
     environment:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD_FILE: /run/secrets/pg_pass
@@ -86,7 +87,7 @@ services:
       POSTGRES_DB: zfindb
       POSTGRES_HOST_AUTH_METHOD: trust
       TZ: ${TZ}
-    shm_size: 1g
+    shm_size: 2g
     ports:
       - "${DOCKER_DB_PORT:-127.0.0.1:5432}:5432"
     volumes:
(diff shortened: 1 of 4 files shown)


cmpich (approved):
Looks good.

PR #1947, opened 2026-08-03
ZIRC: Refactor to cleaner code -- replace hand-written Options "withers" with Lombok @With

rtaylorzfin:
The Options record construction is hard to reason about because it takes 19 parameters. This uses lombok to allow creating Options records using the builder pattern.  It means options.widget(...) calls become options.withWidget(...) calls, but overall I think it's much cleaner. The main file that is improved is Options.java


rtaylorzfin:
Closing as the same refactor exists in #1954

PR #1948, opened 2026-08-04
Liquibase Logging - Reduce noisy output

rtaylorzfin:
Liquibase is writing output about every single SQL file it encounters even if it has already been run. This PR fixes that so we only see the new SQL files.


cmpich (approved):
Nice.

